PR #65, opened 2019-07-02
Updating vsce npm package

jalissia:
This PR updates the vsce package to the latest (1.64.0) to fix the npm audit issue https://nodesecurity.io/advisories/786 .

* Rolling vsce to version 1.64.0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "vscode-edge-devtools",
-    "version": "0.1.0",
+    "version": "1.0.0",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -2229,6 +2049,12 @@
...
             "integrity": "sha1-ji1geottef6IC1SLxYzGvrKIxPM="
         },
+        "didyoumean": {
+            "version": "1.2.1",
+            "resolved": "https://registry.npmjs.org/didyoumean/-/didyoumean-1.2.1.tgz",
+            "integrity": "sha1-6S7f2tplN9SE1zwBcv0eugxJdv8=",
+            "dev": true
+        },
         "diff": {
             "version": "3.3.1",
```
jackjlynch: Doesn't this have to be >= 3.5.0 to fix the issue?

jalissia: You make a good point. This was the result of `npm audit fix` which seemed to only update vsce.

jalissia: Ah I see, i was fixing a different advisory. There isn't an easy fix for the diff one.

I updated the PR description with the actual issue, and I will need to open a new PR to fix this diff one. 

Sorry about that.

Issue #55, opened 2019-06-21
Links from Elements tool don't open up my CSS file


zoherghadyali:
I'm working on this project: https://github.com/googlecodelabs/your-first-pwapp and I have a CSS file at Public/styles/inline.css. I use this launch.json config:

```json
        {
            "type": "vscode-edge-devtools.debug",
            "request": "launch",
            "name": "Launch Microsoft Edge and open the Elements tool",
            "url": "http://localhost:8080",
        }
```

In the Styles pane, I see the correct file and line number. But when I click on it, nothing happens.

![image](https://user-images.githubusercontent.com/34316555/59954594-6715b280-943a-11e9-9a47-9f482f9b1362.png)

Please make it so that when I click on a link in the Elements tool, it opens the file and navigates to the line number in VS Code, like it does in the browser Elements tool.

PR #72, opened 2019-07-12
Adding placeholder code for openInEditor function

jalissia:
This PR is the first step in the feature to open links from the Elements tool in the VS Code editor (see #55). It adds patch code so that the Elements tool will call into the new openInEditor() function when a user clicks a revealer link. The openInEditor() function sends a message to the extension which is where we will actually open the file. The actual open will be implemented in a later PR.

* Added patch for calling openInEditor from the Elements tool
* Added plumbing for passing request to the extension
* Added placeholder for actual opening functionality
* Updated tests

### DIFF
--- a/src/common/webviewEvents.ts
+++ b/src/common/webviewEvents.ts
@@ -1,10 +1,11 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-export type WebviewEvent = "getState" | "getUrl" | "ready" | "setState" | "telemetry" | "websocket";
+export type WebviewEvent = "getState" | "getUrl" | "openInEditor" | "ready" | "setState" | "telemetry" | "websocket";
 export const webviewEventNames: WebviewEvent[] = [
     "getState",
     "getUrl",
+    "openInEditor",
     "ready",
     "setState",
     "telemetry",
--- a/src/devtoolsPanel.ts
+++ b/src/devtoolsPanel.ts
@@ -49,6 +49,7 @@ export class DevToolsPanel {
         this.panelSocket.on("getState", (msg) => this.onSocketGetState(msg));
         this.panelSocket.on("setState", (msg) => this.onSocketSetState(msg));
         this.panelSocket.on("getUrl", (msg) => this.onSocketGetUrl(msg));
+        this.panelSocket.on("openInEditor", (msg) => this.onSocketOpenInEditor(msg));
 
         // Handle closing
         this.panel.onDidDispose(() => {
@@ -166,6 +167,10 @@ export class DevToolsPanel {
         }
 
         encodeMessageForChannel((msg) => this.panel.webview.postMessage(msg), "getUrl", { id: request.id, content });
+    }
+
+    private async onSocketOpenInEditor(message: string) {
+        // TODO: Parse message and open the requested file
     }
 
     private update() {
--- a/src/host/polyfills/simpleView.test.ts
+++ b/src/host/polyfills/simpleView.test.ts
@@ -2,11 +2,31 @@
 // Licensed under the MIT License.
 
 describe("simpleView", () => {
+    it("revealInVSCode calls openInEditor", async () => {
+        const apply = await import("./simpleView");
+        const expected = {
+            columnNumber: 0,
+            lineNumber: 0,
+            uiSourceCode: {
+                _url: "http://bing.com",
+            },
+        };
+        const mockOpen = jest.fn();
+        (global as any).InspectorFrontendHost = {
+            openInEditor: mockOpen,
+        };
+
+        await apply.revealInVSCode(expected, false);
+
+        expect(mockOpen).toHaveBeenCalled();
+    });
+
     it("applyCommonRevealerPatch correctly changes text", async () => {
         const apply = await import("./simpleView");
         const result = apply.applyCommonRevealerPatch(
             "Common.Revealer.reveal = function(revealable, omitFocus) { // code");
-        expect(result).toEqual("Common.Revealer.reveal = function() { Promise.resolve(); return; // code");
+        expect(result).toEqual(
+            expect.stringContaining("Common.Revealer.reveal = function revealInVSCode(revealable, omitFocus) {"));
     });
 
     it("applyInspectorViewPatch correctly changes text", async () => {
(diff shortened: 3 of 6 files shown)


src/devtoolsPanel.ts
```diff
@@ -168,6 +169,10 @@ export class DevToolsPanel {
         encodeMessageForChannel((msg) => this.panel.webview.postMessage(msg), "getUrl", { id: request.id, content });
     }
 
+    private async onSocketOpenInEditor(message: string) {
+        // TODO: Parse message and open the requested file
```
bgoddar: nit: is there a feature request to link this to?

src/host/polyfills/simpleView.test.ts
```diff
@@ -2,11 +2,31 @@
...
+        expect(mockOpen).toHaveBeenCalled();
+    });
+
     it("applyCommonRevealerPatch correctly changes text", async () => {
         const apply = await import("./simpleView");
         const result = apply.applyCommonRevealerPatch(
             "Common.Revealer.reveal = function(revealable, omitFocus) { // code");
-        expect(result).toEqual("Common.Revealer.reveal = function() { Promise.resolve(); return; // code");
+        expect(result).toEqual(
+            expect.stringContaining("Common.Revealer.reveal = function revealInVSCode(revealable, omitFocus) {"));
```
bgoddar: I'm a little confused on this, can you help explain what is happening? I would think expect.stringContaining() to return a boolean, and I thought result would be a string, so how would they be equal? Or is result a boolean as well?

src/host/polyfills/simpleView.ts
```diff
@@ -1,10 +1,34 @@
...
+export function revealInVSCode(revealable: IRevealable | undefined, omitFocus: boolean) {
+    if (revealable && revealable.uiSourceCode && revealable.uiSourceCode._url) {
+        InspectorFrontendHost.openInEditor(
+            revealable.uiSourceCode._url,
+            revealable.lineNumber,
+            revealable.columnNumber,
+        );
+    }
+
+    return Promise.resolve();
```
johnemau: nit: make function async and remove return.

src/host/polyfills/simpleView.test.ts
jalissia: These are special Jest matchers that can be used instead of a literal value. See: https://jestjs.io/docs/en/expect.html#expectstringcontainingstring

You can read it as "expect the result to equal a string containing blah"

src/common/webviewEvents.ts
```diff
@@ -1,10 +1,11 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-export type WebviewEvent = "getState" | "getUrl" | "ready" | "setState" | "telemetry" | "websocket";
+export type WebviewEvent = "getState" | "getUrl" | "openInEditor" | "ready" | "setState" | "telemetry" | "websocket";
```
vidorteg: Why not using websocket to pass the message?

jalissia: This is the type of message being sent. The websocket type is used for when we are passing the devtools CDP json payload that needs to be forwarded over the websocket. 
The new 'openInEditor' type will only be used to tell the extension to open an editor.

src/host/polyfills/simpleView.ts
```diff
@@ -1,10 +1,34 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+import ToolsHost from "../toolsHost";
```
vidorteg: Is there a way to fail te build if some of this regex do not match? Otherwise it will break some features without people noticing if the source code change

src/host/polyfills/simpleView.ts
```diff
@@ -1,10 +1,34 @@
...
+    }
+
+    return Promise.resolve();
+}
+
 export function applyCommonRevealerPatch(content: string) {
     return content.replace(
         /Common\.Revealer\.reveal\s*=\s*function\(revealable,\s*omitFocus\)\s*{/g,
-        "Common.Revealer.reveal = function() { Promise.resolve(); return;");
+        `Common.Revealer.reveal = ${revealInVSCode.toString().slice(0, -1)}`);
```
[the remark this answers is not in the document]
jalissia: It's a hacky way of putting the revealInVSCode function into the devtools js. I'm tostring()-ing the function and stripping off the last end brace, so that I can prepend it into the script.

Patching the code like this in the post build step is not a great way to do it, but right now it does mean that you don't need a custom build of the Edge devtools to build/contribute to the extension. Ideally we would make such changes to the actual product code behind some sort of hosting flag (which the extension turns on by default). But for now this is the mechanism we're using.

src/host/polyfills/simpleView.ts
jalissia: There probably is, but it requires the build to pull down and build the opensource version of Edge in order to test against the elements code. This is a bunch of work to do for the CI system. I can open an issue for it though.

jalissia: Added #73

src/devtoolsPanel.ts
jalissia: The bug is referenced in the description #55, which is tracking the completion of this function

src/host/polyfills/simpleView.ts
jalissia: Due to the post build patching mechanism we are using in this extension (see other comments in this PR) I need to return a promise here.

src/host/polyfills/simpleView.test.ts
bgoddar: That makes sense, my question is: won't expect.stringContaining() return a promise that resolves to pass/fail? and then you are expecting result (which i assume is a string) to equal that promise? Or does expect.stringContaining() actually return the string instead of pass/fail?

jalissia: I'm not 100% sure how the internals of Jest are implemented, but stringContaining() is not a comparison function like the ones that start with 'to'. It actually returns an object (a jest asymmetric matcher) which Jest treats like a literal value when it is chained together with the toEqual call. This lets Jest perform the fuzzy matching logic with the actual value in `result`.

bgoddar: Cool! Thanks for explaining that.

PR #75, opened 2019-07-16
Adding sourcemap related settings ready for opening links in the editor

jalissia:
This PR is the second step in the feature to open links from the Elements tool in the VS Code editor (see #55). It adds a number of settings that can be specified by the user in either their `launch.json` or in the extension settings (for when attach/launch is done using the side panel). The actual opening of the files in the Editor will be done in a follow up PR.

The source map settings match those used by the debugger for Edge extension and will be used to map the runtime source locations reported by the browser, such as `http://localhost:8080/app/file.js` to files that exist on disk, such as `build/out/app/file.ts`.

A new function has been added to get the various runtime settings based on the order of presidence:
launch.json > extension settings > default values

* Added new sourcemap settings to `package.json`
* Updated readme
* Added new getRuntimeConfig() function
* Added unit tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ To add a new debug configuration, in your `launch.json` add a new debug config w
 * `useHttps`: By default the extension will search for attachable instances using the `http` protocol. Set this to true if you are hosting your web app over `https` instead.
 * `sourceMaps`: By default, the extension will use sourcemaps and your original sources whenever possible. You can disable this by setting `sourceMaps` to false.
 * `pathMapping`: This property takes a mapping of URL paths to local paths, to give you more flexibility in how URLs are resolved to local files. `"webRoot": "${workspaceFolder}"` is just shorthand for a pathMapping like `{ "/": "${workspaceFolder}" }`.
-* `sourceMapPathOverrides`: A mapping of source paths from the sourcemap, to the locations of these sources on disk. See [Sourcemps](#sourcemaps) for more information
+* `sourceMapPathOverrides`: A mapping of source paths from the sourcemap, to the locations of these sources on disk. See [Sourcemaps](#sourcemaps) for more information
 
 #### Sourcemaps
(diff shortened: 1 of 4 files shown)


src/utils.ts
```diff
@@ -28,6 +28,20 @@ export interface IUserConfig {
     port: number;
     useHttps: boolean;
     userDataDir: string | boolean;
+    webRoot: string;
```
bgoddar: This is marked as optional in the readme. Should it be optional here? Or is this a case where if the user doesn't provide the webRoot you will fall back to a default and this will always be defined?

jalissia: Yes, this is not set as optional here as the interface specifies all options. In cases where we expect this to be optional the interface is used with Partial<IUserConfig>, and in the other cases it gets a value based on the precedence launch.json > extension setting > default value.

README.md
```diff
@@ -67,7 +69,36 @@ To add a new debug configuration, in your `launch.json` add a new debug config w
...
+
+#### Sourcemaps
+The elements tool uses sourcemaps to correctly open original source files when you click links in the UI, but sometimes the sourcemaps aren't generated properly and overrides are needed. In the config we support `sourceMapPathOverrides`, a mapping of source paths from the sourcemap, to the locations of these sources on disk. Useful when the sourcemap isn't accurate or can't be fixed in the build process.
+
+The left hand side of the mapping is a pattern that can contain a wildcard, and will be tested against the `sourceRoot` + `sources` entry in the source map. If it matches, the source file will be resolved to the path on the right hand side, which should be an absolute path to the source file on disk.
+
+A few mappings are applied by default, corresponding to some common default configs for Webpack and Meteor:
+```json
+// Note: These are the mappings that are included by default out of the box, with examples of how they could be resolved in different scenarios. These are not mappings that would make sense together in one project.
```
bgoddar: nit: It seems odd to me that we are spelling all of these defaults out specifically here, and commenting in them (which makes it invalid json if anyone were to try and copy/modify it for their own use). Not blocking at all, just wondering if instead we just link to where they are defined? Or maybe just list one example?

jalissia: I've reworded the default mappings section so that this is valid json and the examples come after

Issue #44, opened 2019-06-12
Elements tool always opens to the right and not where I last placed it


paulgildea:
When I open a new instance of the Elements tool it always opens a new tab to the right of my current tab. It doesn't remember the last place I put it and open it there.
![image](https://user-images.githubusercontent.com/33229964/59390127-342a3b00-8d25-11e9-9dd1-5e2906788476.png)

vidorteg:
What is the expected behavior in here in the following scenario:

VSCode:
| Tab 1 | Tab 2| Tab 3| Devtools extension | Tab 5|

*close and open devtools*

Possible outcomes:

1. Devtools extension opens in "Tab 4"
1. Devtools extension opens next to the active tab

IMO number 2 is the best option as it shows a consistent behavior with other extensions.

codepo8:
Yes, I also opt for 2.

codepo8:
Can we set option 2?

PR #76, opened 2019-07-16
Changes the way the devtools panel is displayed.

vidorteg:
*Summary*

This PR modifies the way the devtools are displayed, changing from creating a new panel beside the existing window or reusing an empty one (or create a new one if no empty panel is found)

*Before and After*
[https://imgur.com/a/Y9Mi3LI](https://imgur.com/a/Y9Mi3LI)
The first image (before) shows how the devtools are displayed when a new panel is shown and an empty panel exists. The second image(after) shows the same scenario but with the proposed fix. 

I think this is related to issue #44 (depending on the outcome of an active discussion there), so even if this does not fixes issue #44 it improves the way the panel is shown, making it more consistent with the experience get from other extensions.


src/devtoolsPanel.ts
```diff
@@ -210,7 +210,7 @@ export class DevToolsPanel {
         context: vscode.ExtensionContext,
         telemetryReporter: Readonly<TelemetryReporter>,
         targetUrl: string) {
-        const column = vscode.ViewColumn.Beside;
+        const column = vscode.ViewColumn.Active;
```
jalissia: What happens when you only have a single editor window (no columns) open and you launch the elements extension? Does it just become the active tab, or does it open in a second column?

vidorteg: It becomes the active tab, no empty panels are created.

jalissia: Do you think the extra column issue is more annoying than the extra steps it would take to see the code and the elements tool side by side, after this change?
I'm not sure which is a better user experience, but most twitter screenshot's i've seen suggest that people are using a split view with the tool which would take more steps to accomplish after this change.
@brendyna @hxlnt - any thoughts?

vidorteg: Good point, I'm not sure about it. Most of the extensions that I've seen run in the active panel, so for consistence this seems like a good change, but as you mention user will need to drag the active code to the side panel if they want to keep a side by side. I'll leave this to @brendyna and @hxlnt

vidorteg:
@codepo8 thoughts on this one?

src/devtoolsPanel.ts
codepo8: I agree with @jalissia here, having an extra step dragging active code seems to outweigh the benefits. Is there a way to have both, as in follow this way when there is no active code only?

vidorteg: We don't have a way to let the user know, but we can potentially do it. I'll abandon this PR for now :) thanks both.

PR #78, opened 2019-07-18
Update to latest npm packages

jalissia:
This PR just updates all the npm dependencies to the latest versions. This should now fix the 2 github/npm security warnings about lodash and diff.

* Fixes WS-2018-0590
* Fixes CVE-2019-10744


package-lock.json
```diff
@@ -1472,12 +1436,23 @@
                 "htmlparser2": "^3.9.1",
                 "lodash": "^4.15.0",
```
jackjlynch: Is this going to be an issue or is npm smart enough that it'll just use the more recent version that everything else requires?

package-lock.json
```diff
@@ -2138,32 +2051,6 @@
                 "inherits": "^2.0.1",
```
jackjlynch: Github won't let me comment on it because it's so far from the nearest change but there's a dependency on diff 3.3.1 on line 1972, is that going to be an issue?

package-lock.json
```diff
@@ -7352,50 +6672,109 @@
...
+            "resolved": "https://registry.npmjs.org/tslint/-/tslint-5.18.0.tgz",
+            "integrity": "sha512-Q3kXkuDEijQ37nXZZLKErssQVnwCV/+23gFEMROi8IlbaBG6tXqLPQJ5Wjcyt/yHPKBC+hD5SzuGaMora+ZS6w==",
             "dev": true,
             "requires": {
-                "babel-code-frame": "^6.22.0",
+                "@babel/code-frame": "^7.0.0",
                 "builtin-modules": "^1.1.1",
                 "chalk": "^2.3.0",
                 "commander": "^2.12.1",
                 "diff": "^3.2.0",
```
jackjlynch: Same question here as with lodash above

package-lock.json
jalissia: The version it installs of lodash is 4.17.14 so it should be ok since the ^ indicates that a higher version is compatible.

package-lock.json
jalissia: Good catch. I'm just going to force the diff version for now, it looks like its tslint that is pulling in the lower version even though it should be fine with 3.5.0. Probably due to tslint being deprecated soon for eslint (https://github.com/palantir/tslint/issues/4534), so we will need to convert at some point.

package-lock.json
jalissia: See above comment.

jalissia: See above comment.

PR #80, opened 2019-07-23
Moving to new vscode package

jalissia:
This PR updates the package.json file to use the new format for VS Code extensions from the 1.36.0 update. You can read more about the package changes here:
https://code.visualstudio.com/updates/v1_36#_splitting-vscode-package-into-typesvscode-and-vscodetest

The PR follows the migration guidance from here: https://code.visualstudio.com/api/working-with-extensions/testing-extension#migrating-from-vscode

* Updated to use the new @types/vscode and vscode-test packages
* Updated to 1.36 VS Code engine version
* Removed diff dev dependency

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
         "dom"
     ],
     "engines": {
-        "vscode": "^1.33.1"
+        "vscode": "^1.36.0"
     },
     "categories": [
         "Debuggers"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -413,7 +413,6 @@
         "watch": "npm run watch-wp",
         "watch-wp": "webpack --watch",
         "lint": "tslint -c tslint.json -p tsconfig.json",
-        "postinstall": "node ./node_modules/vscode/bin/install",
```
joselea: was this remove intentional?

jalissia: Yes see: https://code.visualstudio.com/api/working-with-extensions/testing-extension#migrating-from-vscode

I'll add this link to the main PR desc. Thanks!

Issue #54, opened 2019-06-21
webRoot attribute in launch.json configuration


zoherghadyali:
So when I use the Debugger for Edge VS Code extension, my launch config looks like: 
```json 
      {
            "type": "edge",
            "request": "launch",
            "version": "canary",
            "name": "Launch Edge",
            "url": "http://localhost:8080",
            "webRoot": "${workspaceFolder}"
        }
```

We use the webRoot attribute to point to the directory that files are served from. It can be either an absolute path or a path using ${workspaceFolder} (the folder open in Code). webRoot is used to resolve urls (like "http://localhost/app.js") to a file on disk (like /Users/me/project/app.js).

Please add the webRoot property so that I don't have to think about different launch.json configs for my 2 Edge extensions.

codepo8:
This is fixed.

PR #83, opened 2019-07-26
Actually opening files in editor from links in the elements tool

### LINKED ISSUE #55: Links from Elements tool don't open up my CSS file

jalissia:
This PR adds the final part of the changes to actually open links in the editor. The onSocketOpenInEditor() function now takes the url from the elements tool (which will be the source mapped url if they are enabled) and calls helpers to convert it to the local workspace path using the config settings. 

Fixes #55, #54

Take a look at this animation to see the results:
elements-links

* Added utils functions to map elements links to vs code file paths (based on the same work in edge debug adapter)
* Added chrome-debug-core package to perform some mapping
* Updated the openInEditor function to actually open the files
* Updated tests and added new ones

### DIFF
--- a/src/devtoolsPanel.ts
+++ b/src/devtoolsPanel.ts
@@ -182,7 +182,15 @@ export class DevToolsPanel {
         });
 
         // Parse message and open the requested file
-        const { column, line, url } = JSON.parse(message) as { column: number, line: number, url: string };
+        const { column, line, url, omitFocus } =
+            JSON.parse(message) as { column: number, line: number, url: string, omitFocus: boolean };
+
+        // If we don't want to focus on the doc and doing so would cause a tab switch ignore it.
+        // This is because just starting to edit a style in the Elements tool with call openInEditor
+        // but if we switch vs code tab the edit will be cancelled.
+        if (omitFocus && this.panel.viewColumn === vscode.ViewColumn.One) {
+            return;
+        }
 
         // Convert the devtools url into a local one
         let sourcePath = url;
--- a/src/devtoolsPanel.ts
+++ b/src/devtoolsPanel.ts
@@ -7,6 +7,7 @@ import * as debugCore from "vscode-chrome-debug-core";
 import TelemetryReporter from "vscode-extension-telemetry";
 import {
     encodeMessageForChannel,
+    IOpenEditorData,
     ITelemetryMeasures,
     ITelemetryProps,
     TelemetryData,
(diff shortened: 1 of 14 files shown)


src/devtoolsPanel.ts
```diff
@@ -179,7 +181,50 @@ export class DevToolsPanel {
...
+        if (uri) {
+            const doc = await vscode.workspace.openTextDocument(uri);
+            vscode.window.showTextDocument(
+                doc,
+                {
+                    preserveFocus: true,
+                    selection: new vscode.Range(line, column, line, column),
+                    viewColumn: vscode.ViewColumn.One,
+                });
+        }
```
bgoddar: Should we maybe have some sort of error message if this doesn't work? I feel like users might be confused if nothing happens.

src/utils.ts
```diff
@@ -353,10 +354,93 @@ export function getRuntimeConfig(config: Partial<IUserConfig> = {}): IRuntimeCon
...
+}
+
+/**
+ * Walk through the list of mappings and find one that matches the sourcePath.
+ * Once a match is found, replace the pattern in the value side of the mapping with
+ * the rest of the path.
+ * @param sourcePath The source path to convert
+ * @param pathMapping The list of mappings from source map to authored file path
+ */
+export function applyPathMapping(
```
bgoddar: I'm surprised there isn't any utilities to do pattern matching/replacing for paths already. But this looks like it should work :)

src/devtoolsPanel.ts
```diff
@@ -179,7 +181,50 @@ export class DevToolsPanel {
             sourceMaps: `${this.config.sourceMaps}`,
         });
 
-        // TODO: Parse message and open the requested file
+        // Parse message and open the requested file
+        const { column, line, url, omitFocus } =
+            JSON.parse(message) as { column: number, line: number, url: string, omitFocus: boolean };
```
robpaveza: This feels like something to have a defined interface for.  Casting to an anonymous interface with four members seems excessive.

src/devtoolsPanel.ts
```diff
@@ -179,7 +181,50 @@ export class DevToolsPanel {
...
+        // Convert the workspace path into a VS Code url
+        let uri: vscode.Uri;
+        try {
+            uri = vscode.Uri.file(sourcePath);
+        } catch {
+            uri = vscode.Uri.parse(sourcePath, true);
+        }
+
+        // Finally open the document if it exists
+        if (uri) {
```
robpaveza: This null check appears unnecessary based on your type which is not nullable.

src/devtoolsPanel.ts
jalissia: Added a new interface

src/devtoolsPanel.ts
jalissia: I guess I wasn't actually making sure that the second parse succeeded, which it might not. Updated the code here (and the type)

src/devtoolsPanel.ts
jalissia: We should, it's just not easy to do that right now without adding some logging logic. Filed #85 to track this work

src/utils.ts
jalissia: I couldn't find any. This is based on the mechanism used by https://github.com/microsoft/vscode-chrome-debug-core/ for sourcemaps.

PR #88, opened 2019-07-31
Enabling localizability of the Elements for Edge extension

vidorteg:
This PR enables the localizability of Elements for Edge extension. 

### After this PR this extension will have:

- Elements UI will be displayed in a supported language, the locale will be autodetected to match VS Code locale.
- locale(s) containing the translations can be provided in a json format, qps-ploc is provided in this PR as an example of the file structure and also to test the localizability on the extension.
- A "Localization" settings to the extension with three values: Enabled, Disabled and qps-ploc.
- A "locale fallback" mechanism, developers can specify which locale(s) and order should be tried before trying the next one.
- **NOTE:** frontend code must support localizability otherwise no visual change will be notable** 
### What this PR is doing?
- Creates the "getStrings" event
- Provides a qps-ploc locale file
- Adds a string provider, this singleton provider:
    - Contains the callback that exposes strings to the UI.
    - Contains the fallback logic
- Modifies the messaging layer, so when the host container is ready it will ask for the loaded resources, if resources are provided they will be exposed by the string provided, otherwise it will be a NO-OP.
- Some tools in the elements panel are whitelisted through the label, as the label can now be localized we need to whitelist the "translated label" too. This is done in the simpleView.ts polyfill
- Modifies utils.ts to load the resources from the specified file (and handles the on/off logic from settings)


src/common/stringsProvider.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    public getStringsCallback(message: any): void {
+        if (!message) {
+            return;
+        }
+
+        if (!this.initialized) {
+            // initialize
+            const data = JSON.parse(message.data);
+            StringsProvider.dtWindow.loadTimeData = { data };
```
lmitchell: Localization.localizedStringsMap

vidorteg:
@lmitchell @jalissia This is ready for review :)

src/devtoolsPanel.ts
```diff
@@ -270,19 +285,26 @@ export class DevToolsPanel {
         context: vscode.ExtensionContext,
         telemetryReporter: Readonly<TelemetryReporter>,
         targetUrl: string,
-        config: IRuntimeConfig) {
+        config: IRuntimeConfig,
+        stringMap: string = "") {
```
jalissia: stringMap is an odd name, I keep thinking is should be a Map<string, string>

src/devtoolsPanel.ts
```diff
@@ -270,19 +285,26 @@ export class DevToolsPanel {
...
-        config: IRuntimeConfig) {
+        config: IRuntimeConfig,
+        stringMap: string = "") {
         const column = vscode.ViewColumn.Beside;
 
         if (DevToolsPanel.instance) {
             DevToolsPanel.instance.panel.reveal(column);
         } else {
-            const panel = vscode.window.createWebviewPanel(SETTINGS_STORE_NAME, SETTINGS_WEBVIEW_NAME, column, {
+            let name = SETTINGS_WEBVIEW_NAME;
```
jalissia: localizedTabTitle would make it clearer what this is actually doing

src/devtoolsPanel.ts
```diff
@@ -270,19 +285,26 @@ export class DevToolsPanel {
...
+            if (stringMap) {
+                const stringMapJSON = JSON.parse(stringMap);
+                name = stringMapJSON[name];
```
jalissia: Should we have a fallback in case the json is missing an entry for the name?

src/extension.test.ts
```diff
@@ -175,6 +176,11 @@ describe("extension", () => {
                     createTelemetryReporter: jest.fn((_: ExtensionContext) => mockTelemetry),
                     fixRemoteWebSocket: jest.fn().mockReturnValue(target),
                     getListOfTargets: jest.fn().mockResolvedValue([target]),
+                    getLocalizedStrings: jest.fn((value) => {
```
jalissia: Could this be simplified to jest.fn().mockResolvedValue(fakeResourceStrings) ?

src/extension.test.ts
```diff
@@ -341,6 +350,11 @@ describe("extension", () => {
                 createTelemetryReporter: jest.fn((_: ExtensionContext) => mockReporter),
                 getBrowserPath: jest.fn().mockResolvedValue("path"),
                 getListOfTargets: jest.fn().mockResolvedValue(null),
+                getLocalizedStrings: jest.fn((value) => {
```
jalissia: ditto

src/extension.ts
```diff
@@ -117,7 +118,8 @@ export async function attach(context: vscode.ExtensionContext, attachUrl?: strin
             // Auto connect to found target
             telemetryReporter.sendTelemetryEvent("command/attach/devtools", telemetryProps);
             const runtimeConfig = getRuntimeConfig(config);
-            DevToolsPanel.createOrShow(context, telemetryReporter, targetWebsocketUrl, runtimeConfig);
+            const stringMap = await getLocalizedStrings(context.extensionPath);
+            DevToolsPanel.createOrShow(context, telemetryReporter, targetWebsocketUrl, runtimeConfig, stringMap);
```
jalissia: Should we just move the getLocalizedStrings call into the createOrShow function?

src/utils.ts
```diff
@@ -444,3 +446,38 @@ export function applyPathMapping(
 
     return sourcePath;
 }
+
+/**
+ * Gets the localized resources that will be displayed in the frontend
+ * @param extensionPath The root folder for the extension path.
+ */
+export async function getLocalizedStrings(extensionPath: string): Promise<string> {
```
jalissia: Missing tests for this function?

src/utils.ts
```diff
@@ -69,6 +70,7 @@ export const SETTINGS_DEFAULT_PATH_OVERRIDES: IStringDictionary<string> = {
     "webpack:///./~/*": "${webRoot}/node_modules/*",
     "webpack:///src/*": "${webRoot}/*",
 };
+export const SETTINGS_LOCALIZATION: string = "Disabled";
```
jalissia: Add the word DEFAULT

src/utils.ts
```diff
@@ -444,3 +446,38 @@ export function applyPathMapping(
...
+
+/**
+ * Gets the localized resources that will be displayed in the frontend
+ * @param extensionPath The root folder for the extension path.
+ */
+export async function getLocalizedStrings(extensionPath: string): Promise<string> {
+    try {
+        const settings = vscode.workspace.getConfiguration(SETTINGS_STORE_NAME);
+        const localizationEnabled: string = settings.get("Localization") || SETTINGS_LOCALIZATION;
+        if (localizationEnabled === "Enabled" || localizationEnabled === "qps-ploc") {
```
jalissia: Add a type definition for the enum options

src/utils.ts
```diff
@@ -444,3 +446,38 @@ export function applyPathMapping(
...
+                    path.join(extensionPath, "resources", "locales", `${locale}.json`));
+                const frontendStringDocument = await vscode.workspace.openTextDocument(resourcePath);
+                if (frontendStringDocument) {
+                    return frontendStringDocument.getText();
+                }
+            }
+        }
+
+        return "";
+    } catch (e) {
```
jalissia: Can we scope down the try/catch to only the part that is actually loading the file?

webpack.config.ts
```diff
@@ -29,6 +29,7 @@ module.exports = [
             filename: "[name].bundle.js",
             path: path.resolve(__dirname, "out/host"),
         },
+        target: "node",
```
jalissia: You can't add this here. The host bundle that this is building is designed to run in the webview, which means it is definitely not running under node! I don't know what you were trying to fix here, but there needs to be another way to do it, since the host bundle needs to be compiled as a browser target not node.

src/devtoolsPanel.ts
vidorteg: it will fallback to the en-us version.

webpack.config.ts
vidorteg: This is the line where I spent most of my time 🍶 , this was to workaround an issue that 'fs' module (used by vscode-nls) has with webpack. As I'm removing vscode-nls this is not needed anymore.

src/common/stringsProvider.ts
```diff
@@ -0,0 +1,68 @@
...
+
+        return;
+    }
+
+    private fallBackInitialization() {
+        // implement custom fallback logic here as follows:
+        // this._fallBackMap.set('en-us', []): if a language is available (json provided)
+        // this._fallBackMap.set('test', ['es-mx']): fallback list for test language
+        this.fallbackMap.set("en-us", []);
+        this.fallbackMap.set("qps-ploc", []);
```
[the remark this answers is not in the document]
vidorteg: The problem here is that fallback logic is not a standard (or even close), it depends a lot on different factors. As an example not all variations of Spanish can fallback to neutral Spanish. The goal with this method is to leave this implementation open to whoever adds the language the idea is that "supported languages" return [] but still you can define a fallback chain e.g 
this.fallbackMap.set("es-vz",["en-ar","es-mx","es-es","es]); 
in this case spanish(Venezuela) will try to fallback to spanish (argentina), spanish(mexico), spanish(spain) and finally to neutral spanish. I'll add some tests to make this more clear.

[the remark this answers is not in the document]
vidorteg: Also open to redesign this logic with a better plan (or to completely remove it and fallback to en-us)

src/common/stringsProvider.ts
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    public static get instance() {
```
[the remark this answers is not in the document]
vidorteg: it could be an instance but as this is a provider I prefer to have it explicitly marked as a singleton to avoid misuse.

src/utils.ts
vidorteg: this one in specific is hard to test as it mostly calls vscode native methods (to get the language and then to load the file), inside the test code it is always mocked.

vidorteg:
I have some tests for stringProvider, @jalissia do you prefer me to include them in this PR, or in a different one?

package.json
```diff
@@ -151,6 +151,21 @@
...
+                },
+                "vscode-edge-devtools.localization": {
+                    "type": "string",
+                    "enum": [
+                        "Enabled",
+                        "Disabled",
+                        "qps-ploc"
+                    ],
+                    "enumDescriptions": [
+                        "Elements for Edge will match the selected locale for VS Code",
```
jalissia: Needs to be Elements for *Microsoft* Edge

src/common/stringProviderConstants.ts
```diff
@@ -0,0 +1,3 @@
+export const englishLocale = "en-us";
```
jalissia: Needs copyright header

package.json
vidorteg: done.

src/common/stringProviderConstants.ts
vidorteg: done.

package.json
```diff
@@ -448,4 +460,4 @@
         "webpack": "4.36.1",
         "webpack-cli": "3.3.6"
     }
-}
+}
```
jalissia: You should change your vs code settings to add a newline to the end of every file, and re-save all these

src/common/webviewEvents.ts
```diff
@@ -10,6 +11,7 @@ export const webviewEventNames: WebviewEvent[] = [
     "setState",
     "telemetry",
     "websocket",
+    "getStrings",
```
jalissia: Move to alphabetical

src/devtoolsPanel.test.ts
```diff
@@ -1,11 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-
-// Allow unused variables in the mocks to have leading underscore
-// tslint:disable: variable-name
-
 import { Disposable, ExtensionContext, WebviewPanel } from "vscode";
 import TelemetryReporter from "vscode-extension-telemetry";
+
```
jalissia: Remove blank line

src/devtoolsPanel.ts
```diff
@@ -1,10 +1,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-
 import * as path from "path";
 import * as vscode from "vscode";
 import * as debugCore from "vscode-chrome-debug-core";
 import TelemetryReporter from "vscode-extension-telemetry";
+
```
jalissia: Ditto

mliao95:
Cleaning out old PRs.  Closing out this PR since it is very dated. We can re-open this PR after updating these changes to work with the tip of branch if some of these changes are still applicable.

FYI @vidorteg

PR #89, opened 2019-08-02
Bubbling up close message from websocket to the extension.

vidorteg:
This PR bubbles up the close message from the websocket to the extension, so when the websocket is closed the extension will be closed too.

### DIFF
--- a/src/common/webviewEvents.ts
+++ b/src/common/webviewEvents.ts
@@ -1,7 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-export type WebviewEvent = "getState" | "getUrl" | "openInEditor" | "ready" | "setState" | "telemetry" | "websocket";
+export type WebviewEvent = "getState" | "getUrl" | "openInEditor" | "ready" | "setState" | "telemetry" | "websocket" |
+                            "close";
 export const webviewEventNames: WebviewEvent[] = [
     "getState",
     "getUrl",
@@ -10,6 +11,7 @@ export const webviewEventNames: WebviewEvent[] = [
     "setState",
     "telemetry",
     "websocket",
+    "close",
 ];
 
 export type WebSocketEvent = "open" | "close" | "error" | "message";
--- a/src/common/webviewEvents.ts
+++ b/src/common/webviewEvents.ts
@@ -1,8 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-export type WebviewEvent = "getState" | "getUrl" | "openInEditor" | "ready" | "setState" | "telemetry" | "websocket" |
-                            "close";
+export type WebviewEvent = "getState" | "getUrl" | "openInEditor" | "ready" | "setState" | "telemetry" | "websocket";
 export const webviewEventNames: WebviewEvent[] = [
     "getState",
     "getUrl",
@@ -11,7 +10,6 @@ export const webviewEventNames: WebviewEvent[] = [
     "setState",
     "telemetry",
     "websocket",
-    "close",
 ];
 
 export type WebSocketEvent = "open" | "close" | "error" | "message";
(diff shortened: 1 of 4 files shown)


src/panelSocket.ts
```diff
@@ -102,11 +102,13 @@ export class PanelSocket extends EventEmitter {
         }
     }
 
-    private onClose() {
+    private onClose(message: { wasClean: boolean; code: number; reason: string; target: WebSocket }) {
```
jalissia: You aren't really using this message param (which is actually an event rather than a message).

Let's just remove it and add a real 'closed' event to the PanelSocket that we can emit here.

src/devtoolsPanel.ts
```diff
@@ -51,7 +51,7 @@ export class DevToolsPanel {
         // Hook up the socket events
         this.panelSocket = new PanelSocket(this.targetUrl, (e, msg) => this.postToDevTools(e, msg));
         this.panelSocket.on("ready", () => this.onSocketReady());
-        this.panelSocket.on("websocket", () => this.onSocketMessage());
+        this.panelSocket.on("websocket", (msg) => this.onSocketMessage(msg));
```
jalissia: Let's add a real closed event for the panelSocket rather than overloading the message event.

src/devtoolsPanel.ts
```diff
@@ -109,8 +109,11 @@ export class DevToolsPanel {
             this.panelSocket.isConnectedToTarget ? "websocket/reconnect" : "websocket/connect");
     }
 
-    private onSocketMessage() {
-        // TODO: Handle message
+    private onSocketMessage(msg: any) {
```
jalissia: Move this into a new handler for close

src/devtoolsPanel.ts
jalissia: Make sure to add a new test with these changes

src/devtoolsPanel.ts
```diff
@@ -57,6 +57,7 @@ export class DevToolsPanel {
         this.panelSocket.on("setState", (msg) => this.onSocketSetState(msg));
         this.panelSocket.on("getUrl", (msg) => this.onSocketGetUrl(msg));
         this.panelSocket.on("openInEditor", (msg) => this.onSocketOpenInEditor(msg));
+        this.panelSocket.on("close", () => this.onSocketClose());
```
jalissia: You are failing a test here because it used to make sure that we were handling all the events, but now you've added a new one it doesn't know about.

Be sure to update that test, and then add a new one which checks that firing 'close' will actually dispose the panel.

src/common/webviewEvents.ts
```diff
@@ -1,7 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-export type WebviewEvent = "getState" | "getUrl" | "openInEditor" | "ready" | "setState" | "telemetry" | "websocket";
+export type WebviewEvent = "getState" | "getUrl" | "openInEditor" | "ready" | "setState" | "telemetry" | "websocket" |
+                            "close";
```
jalissia: This should only be used for messages that go between the devtools and the extension. 'close' is just being emitted on the extension side only in response to the real websocket closing down.
Instead of adding this here to fix the 'listens for all the emit events' test, you should update the test to just check that the number of events it has hooked is +1 // for the close event. And then add a check that ```expect(hookedEvents).toContain("close")```

src/devtoolsPanel.test.ts
```diff
@@ -161,10 +161,14 @@ describe("devtoolsPanel", () => {
             const dtp = await import("./devtoolsPanel");
             dtp.DevToolsPanel.createOrShow(context, mockTelemetry, "", mockRuntimeConfig);
 
-            expect(mockPanelSocket.on).toHaveBeenCalledTimes(Object.keys(webviewEventNames).length);
+            expect(mockPanelSocket.on).toHaveBeenCalledTimes(Object.keys(webviewEventNames).length + 1);
```
jalissia: Add a small comment to explain the +1 E.g. // +1 for 'close' event

PR #96, opened 2019-09-04
Adding support for auto attaching Elements when debugging Edge

### LINKED ISSUE #93: `Elements for Microsoft Edge: Attach to a target` command does nothing
I've been trying to figure out how to attach Elements to my running browser instance launched by Debugger for Edge.

The existing docs aren't helping and the "Elements for Microsoft Edge: Attach to a target" command does nothing except put up an unproductive edit field.  I expect that, according to the docs, it will put up a dropdown allowing me to select the browser instance to which to attach.

jalissia:
This PR adds the ability for the Elements for Edge extension to detect when the user starts debugging with the [Debugger for Microsoft Edge](https://marketplace.visualstudio.com/items?itemName=msjsdiag.debugger-for-edge) extension. When debugging is started, the Elements tool will attempt to attach to the same instance using the same config settings for the debugger after a specified timeout (since we can't easily detect when the browser has finished loading). This will cause the Elements tool to open up when you start debugging instead of needing a second manual attach step.

* Fixes #93 
* Added registerDebugConfigurationProvider call for `edge` to detect debugging start and auto attach the Elements tool
* Added new settings to turn off auto attach and specify a new timeout
* Added code to map configs from Edge debugger to Elements and new urlFilter option
* Updated readme
* Updated tests

elements-debugger-pr


README.md
```diff
@@ -18,6 +18,8 @@ A VS Code extension that allows you to use the browser's Elements tool from with
 * Side Bar view for listing all the debuggable targets, including tabs, extensions, service workers, etc.
 * Fully featured Elements tool with views for HTML, CSS, accessibility and more.
 * Screen-casting feature to allow you to see your page without leaving VSCode.
+* Go direct to the line/column for source files in your workspace when clicking on a link inside the Elements tool.
```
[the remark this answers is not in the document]
jalissia: Done.

package.json
```diff
@@ -151,6 +151,16 @@
...
+                },
+                "vscode-edge-devtools.autoAttachViaDebuggerForEdge": {
+                    "type": "boolean",
+                    "description": "Auto attach the Elements for Microsoft Edge extension when you launch a browser with the Debugger For Microsoft Edge debug adapter extension",
+                    "default": true
+                },
+                "vscode-edge-devtools.debugAttachTimeoutMs": {
+                    "type": "number",
+                    "description": "How many milliseconds to wait before attaching the Elements tool after launching a debugger",
+                    "default": 3000
```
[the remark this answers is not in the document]
jalissia: I'm hoping so, if we get a lot of issues we can up the default value. I've added a section to the readme that talks a bit about this.

brendyna:
> Seems reasonable to me. nit about wording and question about the timeout time.
> 
> One question I do have: are we sure we want to enable by default? I think that would help with discoverability and with ease of use for the use case we are envisioning, I'm just wondering if users may be surprised or unhappy if all of a sudden elements starts auto-launching after they update the extension.

That's a fair concern. I wonder if they'd be less confused or bothered given you have to have both extensions installed for the behavior to work? We can publicize how to turn it off, as well, and if there's strong backlash consider turning it off by default.

README.md
```diff
@@ -18,6 +18,8 @@ A VS Code extension that allows you to use the browser's Elements tool from with
 * Side Bar view for listing all the debuggable targets, including tabs, extensions, service workers, etc.
 * Fully featured Elements tool with views for HTML, CSS, accessibility and more.
 * Screen-casting feature to allow you to see your page without leaving VSCode.
+* Go directly to the line/column for source files in your workspace when clicking on a link inside the Elements tool.
```
brendyna: Or clicking on an individual CSS rule, right? (Or is that only once the file is opened?)

package.json
brendyna: Why'd we go with 3000 here?

jalissia: I decided to change the attach model from a one time try after 3000ms, to a polling model similar to what the Edge debugger is doing. So the default is now 10seconds, but it is the amount of time to keep trying now, rather than the time to wait.

PR #101, opened 2019-10-08
Updating icon to be seen on white background

jalissia:
This PR updates the icon from white to gray so that it is visible in the various places it is used by vs code:

icon-change

* Updated icon
* Updated version number and changelist


codepo8:
Just when I had to re-start my phone for SIM 2FA Fabio already comes to the rescue

PR #102, opened 2019-10-14
Modifying polyfills to comply with latest changes in the code.

vidorteg:
Modifying several polyfills to work with the latest version of the code, also leaving existing code for backwards compatibility

### elementsTreeOutline.js
Renaming the target file to elementsTreeOutline.css

### stylesSectionTree.js
Renaming the target file to stylesSectionTree.css

### themes/dark.css
Applyting existing transformations to dark theme.

### TabbedPaned.js:
 applySelectTabPatch method is changing it's signature. This PR adds the logic to handle the new signature as well as providing backward compatibility with the former signature.

Changes on TabbedPane.js
https://cs.chromium.org/chromium/src/third_party/blink/renderer/devtools/front_end/ui/TabbedPane.js?q=tabbedPane.js+-f:out&sq=package:chromium&g=0&l=1

### DIFF
--- a/postBuildStep.ts
+++ b/postBuildStep.ts
@@ -108,14 +108,14 @@ async function patchFilesForWebView(toolsOutDir: string) {
     await patchFileForWebView("ui/InspectorView.js", toolsOutDir, false, [applyInspectorViewPatch]);
     await patchFileForWebView("ui/TabbedPane.js", toolsOutDir, false, [applySelectTabPatch]);
     await patchFileForWebViewWithLegacySupport(
-                              "elements/elementsTreeOutline.js",
                               "elements/elementsTreeOutline.css",
+                              "elements/elementsTreeOutline.js",
                               toolsOutDir,
                               false,
                               [applyPaddingInlineCssPatch]);
     await patchFileForWebViewWithLegacySupport(
-                              "elements/stylesSectionTree.js",
                               "elements/stylesSectionTree.css",
+                              "elements/stylesSectionTree.js",
                               toolsOutDir,
                               false,
                               [applyPaddingInlineCssPatch]);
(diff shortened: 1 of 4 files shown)


src/host/polyfills/simpleView.ts
```diff
@@ -94,3 +105,32 @@ export function applyInspectorCommonCssPatch(content: string, isRelease?: boolea
         `${css}${separator} $1`,
     );
 }
+
+export function applyInspectorDarkCssPatch(content: string, isRelease?: boolean) {
```
vidorteg: This can replace applyInspectorCommonCssPatch also, but I prefer to leave it separate as multipurpose this behavior will probably lead to some errors in the future.  Thoughts?

postBuildStep.ts
```diff
@@ -140,4 +153,23 @@ function isDirectory(fullPath: string) {
     }
 }
 
+async function patchFileForWebViewWithLegacySupport(
```
jalissia: I think we can remove this function and just assume we always build with the latest version of the tools. We probably wouldn't be using this patching mechanism if we wanted to host older versions of the tools in the webview anyway.

src/host/polyfills/simpleView.ts
```diff
@@ -71,9 +71,20 @@ export function applySelectTabPatch(content: string) {
         return `id !== '${v}'`;
     }).join(" && ");
 
+    // v1Pattern represent legacy version of the API, v2 represents the latest
+    // method signature.
+    const v1Pattern = /selectTab\(id,\s*userGesture\)\s*{/g;
+    const v2Pattern = /selectTab\(id,\s*userGesture,\s*forceFocus\)\s*{/g;
```
jalissia: Same here, lets just assume we always build with the latest.

PR #107, opened 2019-11-14
Fixing issues caused by chromium devtools repository migration.

vidorteg:
Chromium repository migrated to a new repository and a new path, this PR address the following issues caused by this:

- Fixes the path for devtools folder and updates postBuildStep with the new path.
- Fixes an issue caused by moving window.Runtime to window.Root.Runtime
- Fixes an issue caused by session storage not being available for the developer tools as VsCode does not exposes it
- Fixes associated tests

### DIFF
--- a/src/host/host.test.ts
+++ b/src/host/host.test.ts
@@ -61,16 +61,16 @@ describe("host", () => {
             expect(mockDevToolsWindow.localStorage).toBeUndefined();
         });
 
-        it("verifies that session storage cannot be modified", async () => {
+        it("emulates session storage", async () => {
             const host = await import("./host");
             host.initialize(mockIframe);
 
             expect(mockDevToolsWindow.sessionStorage).toBeDefined();
-            const previousSessionStorage = mockDevToolsWindow.sessionStorage;
-            const setter = Object.getOwnPropertyDescriptor(mockDevToolsWindow, "sessionStorage");
-            expect(setter).toBeDefined();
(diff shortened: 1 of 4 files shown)


src/host/host.test.ts
```diff
@@ -59,6 +60,18 @@ describe("host", () => {
             setter!.set!("some new value");
             expect(mockDevToolsWindow.localStorage).toBeUndefined();
         });
+
+        it("verifies that session storage cannot be modified", async () => {
```
jalissia: Probably would be better to test that session storage actually still works, rather than cannot be modified, something like: 
```
        it("emulates session storage", async () => {
            const host = await import("./host");
            host.initialize(mockIframe);

            expect(mockDevToolsWindow.sessionStorage).toBeDefined();

            const expectedKey = "key";
            const expectedVal = "value";
            (mockDevToolsWindow as any).sessionStorage[expectedKey] = expectedVal;
            expect(mockDevToolsWindow.sessionStorage[expectedKey]).toEqual(expectedVal);
        });
```

src/host/toolsResourceLoader.ts
```diff
@@ -3,6 +3,9 @@
 
 import { encodeMessageForChannel } from "../common/webviewEvents";
 
+export interface IRoot {
```
jalissia: Let's move this into `host.ts` and just use it from there since it is the only thing that really needs it.

PR #109, opened 2019-11-15
Remove iframe indirection and update content security policy

### LINKED ISSUE #108: Elements extension doesn't work with VS Code update 1.40
It looks like the move to electron 6 (chromium 76) has changed some of the security concepts for the VS Code webview that we are using.

The extension fails to render any content in the webview due to cross domain issues that were not present in the previous version of electron's chromium.

jalissia:
This PR updates the extension to be compatible with VS Code v1.40. The upgrade to chromium 76 in the latest version of VS Code changed the way the vscode-resource:// protocol is handled by the VS Code WebView. During postmessage() the `origin` property is now reported as `"null"` so we need to handle the messaging redirection using another mechanism. Also the content security policy now treats our vscode-resource:// urls used on the WebView as different domains, so I added code to modify the inspector.html in place to remove a layer of indirection causing the issue.

* Update content security policy to include `vscode-resource://`
* Remove devtools.html iframe indirection
* Update messaging handler to account for null origin
* Add tests
* Fixes #108


src/devtoolsPanel.ts
```diff
@@ -266,7 +266,7 @@ export class DevToolsPanel {
                 <script src="${scriptUri}"></script>
             </head>
             <body>
-                <iframe id="host" frameBorder="0" src="${htmlUri}"></iframe>
+                <iframe id="host" frameBorder="0" src="${htmlUri}?ws=trueD&experiments=true&edgeThemes=true"></iframe>
```
vidorteg: why is this needed?

src/host/messaging.ts
```diff
@@ -13,7 +13,14 @@ export function initializeMessaging() {
     });
 
     window.addEventListener("message", (messageEvent) => {
-        if (messageEvent.origin === "vscode-resource://") {
+        // Both windows now have a "null" origin so we need to distiguish direction based on protocol,
+        // which will throw an exception when it is from the devtools x-domain window.
+        let sendToDevTools = false;
```
vidorteg: Wouldn't rely on this approach will be super expensive? (relying on exception handling as logic for the code)

jalissia: Unfortunately I can't think of another way around this. Since the origin is now "null" for both the data:// and vscode-resource:// origins, the only way to tell which it is, is to check something on the window (source) object. But that will throw an exception when it is pointing at the iframe. This is only true for VSCode > 140 which is why it actually compares the protocol so that in <140 it will not throw and still know the direction.

src/devtoolsPanel.ts
jalissia: This used to be in the devtools.html file where it linked inspector.html as an iframe. Since that file no longer exists we need to move the query params to where inspector.html is now loaded, which is here.

src/host/messaging.ts
vidorteg: can we create an issue and link that in here with a comment, I would like to come back in the future, were hopefully there is a fix for this and address the issue.

jalissia: Opened #110

PR #116, opened 2019-12-12
Bumping dependencies version.

vidorteg:
Updating indirect dependencies.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "vscode-edge-devtools",
-    "version": "1.0.4",
+    "version": "1.0.6",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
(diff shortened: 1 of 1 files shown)


jalissia:
Shouldn't there be a package.json file change not just the lock file?

vidorteg:
AFAIK package.json just tracks direct dependencies, while package.json.lock tracks indirect dependencies. In this we just needed to update indirect dependencies.

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "vscode-edge-devtools",
-    "version": "1.0.4",
+    "version": "1.0.6",
```
jalissia: Shouldn't this be in the package.json file?

jalissia: Shouldn't this be in the package.json file?

vidorteg: it is 😃 , this PR bumped the package.json version:
https://github.com/microsoft/vscode-edge-devtools/commit/f7ab8a7a9d5b14a5732c631f4ab8f290ea154e00

jalissia: Oh i see. So it was actually my fault oops!

PR #136, opened 2020-05-06
Adding an extension setting that will  toggle network panel on and off

vidorteg:
This PR introduces a setting that allows to control whether the "Network" panel should be displayed or not. 

New setting:
 image

When _checked_ the network panel will be displayed.

### DIFF
--- a/src/host/polyfills/simpleView.ts
+++ b/src/host/polyfills/simpleView.ts
@@ -156,6 +156,7 @@ export function applyAppendTabPatch(content: string) {
     if (content.match(injectionPoint)) {
         content = content.replace(
             injectionPoint,
+            // points to the getApprovedTabs, then take the out the "function" from the string (slice(9))
             `${getApprovedTabs.toString().slice(9)};
             appendTab(id, tabTitle, view, tabTooltip, userGesture, isCloseable, index) {
                 let patchedCondition = ${condition};
--- a/src/host/polyfills/simpleView.ts
+++ b/src/host/polyfills/simpleView.ts
@@ -26,6 +26,10 @@ export function revealInVSCode(revealable: IRevealable | undefined, omitFocus: b
     return Promise.resolve();
 }
 
+export function getApprovedTabs(callback: () => void) {
+    InspectorFrontendHost.getApprovedTabs(callback);
+}
+
 export function applyCommonRevealerPatch(content: string) {
     const pattern = /let reveal\s*=\s*function\(revealable,\s*omitFocus\)\s*{/g;
     if (content.match(pattern)) {
@@ -123,7 +127,7 @@ export function applySetTabIconPatch(content: string) {
 export function applyAppendTabPatch(content: string) {
     // The appendTab function chooses which tabs to put in the tabbed pane header section
     // showTabElement and selectTab are only called by tabs that have already been appended via appendTab.
-    const allowedTabs = [
+    const elementsTabs = [
         "elements",
         "Styles",
         "Computed",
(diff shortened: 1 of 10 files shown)


mliao95:
Sorry for the delay in reviewing.  I'll take a closer look next week, but there are a couple high-level points I wanted to discuss.

The first is regarding the main header tab bar.  Before merging the network tool, this was hidden.  Do you think it is worth re-hiding it when the Network tab is hidden?
image
Fix: Re-implement this function and apply it only when the network tool is disabled: https://github.com/microsoft/vscode-edge-devtools/blob/8abd6ee7a93bc61da4c7ed3f4aedbf85281a5067/src/host/polyfills/simpleView.ts#L104

The next is applying the applyHandleAction patch to InspectorView.js.  Applying the applyHandleAction patch to InspectorView.js will disable open/closing the drawer using "ESC" and changing tabs using "CTRL + ]" and "CTRL + [".  This should be disabled when the network tool is hidden as both the drawer toggle and tab change is useless (changing tabs doesn't work for the styles pane tabs, just the main toolbar)

Fix: Apply applyHandleAction patch to InspectorView.js here:
https://github.com/microsoft/vscode-edge-devtools/blob/master/postBuildStep.ts#L108

src/host/polyfills/simpleView.ts
```diff
@@ -121,15 +125,58 @@ export function applySetTabIconPatch(content: string) {
...
+            `appendTabOverride(id, tabTitle, view, tabTooltip, userGesture, isCloseable, index) {`);
+    } else {
+        return null;
+    }
+
+    // We then replace with the verifications itself.
+    if (content.match(injectionPoint)) {
+        content = content.replace(
+            injectionPoint,
+            `${getApprovedTabs.toString().slice(9)};
```
mliao95, line 159: NIT: I see now that here is where we are injecting the getApprovedTabs function.  Might be worth leaving a comment here explaining why we are injecting "${getApprovedTabs.toString().slice(9)}" and what it is.

src/host/toolsHost.ts
```diff
@@ -80,11 +80,17 @@ export default class ToolsHost {
...
 
+    public getApprovedTabs(callback: () => void) {
+        const id = this.getHostCallbacksNextId++;
+        this.getHostCallbacks.set(id, callback);
+        encodeMessageForChannel((msg) => window.parent.postMessage(msg, "*"), "getApprovedTabs", {id});
+    }
+
     public onMessageFromChannel(e: WebviewEvent, args: string): boolean {
         switch (e) {
             case "getState": {
```
mliao95, line 91: Could you go into a little more detail into why we want to make the distinction between "getState" and "getHost"?  It appears that we care enough to change the names of the variables in this file, but not enough to create a new webview event.  To me, it just seems odd that now we're firing "fireGetHostCallback" on "getState."

Happy to keep this change, but if we do decide to change it, it should also be reflected in the toolsHost.test.ts file as well in places such as here: https://github.com/microsoft/vscode-edge-devtools/blob/master/src/host/toolsHost.test.ts#L44

If we want to make the distinction between "getState" vs "getHost", do you think we should make the same changes for the

vidorteg: Sure, 
getState is a webview event created to retrieve the "state" for the devtools, meaning the preferences that devtools require for initialization. To achieve this it uses a synchronization mechanism:

    private **getStateCallbacks**...

While I was coding the solution for the network I realized that my code was looking pretty similar to this, so I generalized the implementation to avoid code repetition. Thus becoming:

    private **getHostCallbacks:**...

GetState is still an event that hints that devtools preferences need to be retrieved, but it will now use the generalized implementation **fireGetHostCallback** to synchronize. Does this answer the question?

mliao95: Yeah that makes sense.  LGTM then.

PR #144, opened 2020-05-19
Fixing sourceMapPathOverrides description

vidorteg:
**Root cause:**
sourceMapPathOverrides is confusing as it has the same text as pathMapping.

**Fix:**
Provide a more accurate description.

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,7 +138,7 @@
                 },
                 "vscode-edge-devtools.sourceMapPathOverrides": {
                     "type": "object",
-                    "description": "A set of mappings to override the locations of sourcemaps files.",
+                    "description": "A set of mappings to override the locations of source map files.",
                     "default": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -138,7 +138,7 @@
                 },
                 "vscode-edge-devtools.sourceMapPathOverrides": {
                     "type": "object",
-                    "description": "A set of mappings for rewriting the locations of source files from what the sourcemap says, to their locations on disk",
+                    "description": "A set of mappings to override the locations of sourcemaps files.",
```
mliao95, line 141: NIT: I think we want to add a space so it is "source maps."

I'm not too familiar with the singular/plural grammar for source maps, but should it be "locations of source map files" or do we want "source maps" to be plural as well?

vidorteg: not an expert either, but "locations of source map files" sounds good to me. 👍

PR #145, opened 2020-05-20
Update Azure Pipelines to Include Edge Files

mliao95:
This PR updates the Azure Pipeline to include Edge files while building and testing and publishes the .vsix artifact.  Since the new pipeline has access to the Edge files, I am failing the build step when we encounter an error.

Original Pipeline Steps:
image

New Pipeline Steps:
image

Original Build Failure Handling:
image

New Build Failure Handling:
image


src/scripts/retrieveDownloadLink.js
```diff
@@ -0,0 +1,19 @@
+function fetchJsonFromUrl(url){
```
vidorteg, line 1: I think this should live outside of _src_, in it's own folder

mliao95: Done.

PR #147, opened 2020-05-22
Converting from browserPath to browserFlavor

vidorteg:
This PR removes the _browserPath setting_ and creates a _browser flavor combo box_ for a similar purpose.

image

**Current Behavior:**
The user can specify (via a input box) the path to a browser version, when no path is specified the extension searches for stable > beta > dev > canary (in that order) until an existing version is found on disk.

**New Behavior**
When default is selected, extension will keep current behavior, otherwise it will use the version specified by the user. 

Test are also modified to use a specific flavor instead of a custom path.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -209,7 +209,7 @@ export function getRemoteEndpointSettings(config: Partial<IUserConfig> = {}): ID
     // We generate a temp directory if the user opted in explicitly with 'true' (which is the default),
     // Or if it is not defined and they are not using a custom browser path (such as electron).
     // This matches the behavior of the chrome and edge debug extensions.
-    const browserPathSet = config.browserFlavor !== "Default";
+    const browserPathSet = config.browserFlavor || "Default";
     let userDataDir: string | boolean | undefined;
     if (typeof config.userDataDir !== "undefined") {
         userDataDir = config.userDataDir;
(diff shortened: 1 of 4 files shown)


mliao95:
Just did a quick first pass and it works well!  One NIT would be to just capitalize the options.  Will take a closer look tomorrow.

vidorteg:
> Just did a quick first pass and it works well! One NIT would be to just capitalize the options. Will take a closer look tomorrow.

done :)

src/utils.ts
```diff
@@ -261,32 +256,25 @@ export function getPlatform(): Platform {
...
-        for (const p of searchPaths) {
-            if (await fse.pathExists(p)) {
-                return p;
-            }
+    const flavor: BrowserFlavor | undefined = config.browserFlavor || settings.get("browserFlavor");
+
+    switch (getPlatform()) {
+        case "Windows": {
+           return await verifyFlavorPath(flavor, "Windows");
+           break;
```
mliao95, line 269: We don't need break statements if we are returning the line before.

src/utils.ts
```diff
@@ -225,7 +220,7 @@ export function getRemoteEndpointSettings(config: Partial<IUserConfig> = {}): ID
         }
     }
 
-    if (userDataDir === true || (typeof userDataDir === "undefined" && !browserPath)) {
+    if (userDataDir === true || (typeof userDataDir === "undefined" && browserPathSet)) {
```
mliao95, line 223: Correct me if I am wrong, but I think we either want to flip the boolean here or in browserPathSet above.

browserPath would return false if the user used default and would return true if the user had an alternate path
browserPathSet would also return false if the browserFlavor was default and return true if the user set it to anything else but default

It looks like these booleans return true/false for default/not default, so we shouldn't remove the "!" from the if statement.

vidorteg: good catch!

src/utils.ts
vidorteg: 😱, how did the linter not catch that one?! nice one Michael.

PR #155, opened 2020-06-02
Deleting CustomElements polyfill

mliao95:
Modifying CustomElements polyfill to return null when no text is replaced. Modifying TextSelection tests to take code from the source file. 

Also changing patch target files and patching contents to work with version 81.0.416


mliao95:
We might not need this patch.  It looks like initially, these patches were used for older versions of webview and would be only applied to versions older than 67: https://github.com/microsoft/vscode-edge-devtools/commit/f73611db5c007334bf195f5af71a250528e7a7c7

In our current code, it is being applied to shell.js and resulting in a no-op due to targeting the wrong file.  After finding the correct files in Edge version 81 and modifying the tests to be compatible with version 81, applying either of the two patches will result in a blank extension.

We should probably re-evaluate if we need this patch since we haven't been using it for a while (since it was applied incorrectly) and the use case that it was created for no longer exists in our code (webview version checking).

Do we know what these patches were used for and if we still need them?

vidorteg:
> We might not need this patch. It looks like initially, these patches were used for older versions of webview and would be only applied to versions older than 67: [f73611d](https://github.com/microsoft/vscode-edge-devtools/commit/f73611db5c007334bf195f5af71a250528e7a7c7)
> 
> In our current code, it is being applied to shell.js and resulting in a no-op due to targeting the wrong file. After finding the correct files in Edge version 81 and modifying the tests to be compatible with version 81, applying either of the two patches will result in a blank extension.
> 
> We should probably re-evaluate if we need this patch since we haven't been using it for a while (since it was applied incorrectly) and the use case that it was created for no longer exists in our code (webview version checking).
> 
> Do we know what these patches were used for and if we still need them?

Not sure why this were created, but we may not need them anymore and if we do need them we can add them back, I prefer not have code that is not being used around. Can this refactoring then be remove them as they are not being used?

PR #156, opened 2020-06-04
inspectorContentPolicy polyfill refactor

mliao95:
Modifying inspectorContentPolicy polyfill to return null when no text is replaced. Modifying inspectorContentPolicy tests to take code from the source file.

### DIFF
--- a/src/host/polyfills/inspectorContentPolicy.ts
+++ b/src/host/polyfills/inspectorContentPolicy.ts
@@ -11,7 +11,7 @@ export function applyContentSecurityPolicyPatch(content: string) {
     }
 
     const metaString = `<meta name="referrer" content="no-referrer">`;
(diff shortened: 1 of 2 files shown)


src/host/polyfills/inspectorContentPolicy.ts
```diff
@@ -2,12 +2,18 @@
...
+    const scriptPattern = /script-src\s*'self'/g;
+    let result;
+    if (content.match(scriptPattern)) {
+        result = content.replace(scriptPattern, `script-src vscode-resource: 'self'`);
+    } else {
+        return null;
+    }
+
+    const metaString = `<meta name="referrer" content="no-referrer">`;
+    if (result.indexOf(metaString) !== -1) {
```
vidorteg, line 14: nit: content.match(...

PR #159, opened 2020-06-05
Fix pipeline: Adding platform args to download script and adding unzipping bash script

mliao95:
# Changes
- updating retrieveDownloadLink to accept an argument for which platform
- Adding scripts for windows users and for the pipeline to easily extract from the .zip file
- unzipEdge.sh works best for Windows users downloading the DevTools .zip
- pipelineUnzipEdge.sh performs the same function with minor changes to work in a linux based environment

### DIFF
--- a/scripts/unzipEdge.sh
+++ b/scripts/unzipEdge.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+{
+    unzip $1 src/out/Release/gen/devtools/* -d $2
+    unzip $1 src/out/Release/gen/devtools/*/* -d $2
+    unzip $1 src/third_party/devtools-frontend/src/front_end/*/* -d $2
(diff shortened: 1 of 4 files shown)


scripts/unzipEdge.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+{
+    unzip $1 src/out/Release/gen/devtools/* -d $2
+    unzip $1 src/out/Release/gen/devtools/*/* -d $2
```
vidorteg, line 4: Could you open an issue to remove support for the old directory support and tag the code that needs to be removed with it?

mliao95: Created: https://github.com/microsoft/vscode-edge-devtools/issues/161

Will link in code once this PR merges

PR #166, opened 2020-06-09
Drafts that enables version 83

vidorteg:


mliao95:
Need to update the pipeline to grab other versions

scripts/retrieveDownloadLink.js
```diff
@@ -1,4 +1,4 @@
-const TARGET_VERSION = '81.0.416.72';
+const TARGET_VERSION = '83.0.478.37';
```
mliao95, line 1: 83.0.478.45 is the one we have on thirdpartysource

vidorteg: thanks @mliao95 .

src/host/polyfills/runtime.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+export function applyRuntimeImportScriptPathPrefixPatch(content: string) {
```
mliao95, line 4: We should do the same refactor on this polyfill where it will return null if the regex does not find a match.

mliao95: Also a test for this function would also be nice.

PR #176, opened 2020-07-17
Updating some missing webview resources

vidorteg:
Issue:
Resources were not loaded, which broke the extension

Fix:
Updating some resources with the latest practices for webview.

Related discussion here:
https://github.com/microsoft/vscode/issues/101562


src/host/messaging.ts
```diff
@@ -18,7 +18,7 @@ export function initializeMessaging() {
         // See: https://blog.mattbierner.com/vscode-webview-web-learnings/
```
mliao95, line 18: You were saying that we should ignore this comment.  Should we remove it if it doesn't apply anymore?

Issue #173, opened 2020-07-01
Create the option to attach to headless browser


vidorteg:
Got some feedback around creating an option to attach to the browser in headless mode and completely rely on screencast. This sounds like a good idea so opening the request for this.

PR #185, opened 2020-08-05
Introducing "headless" browser launch mode

vidorteg:
This PR introduces a new Setting called:
- Headless (off by default)

Which allow the user to use the extension in headless mode. The headless browser tabs will be closed when the associated visual studio code session ends.

Tests are also provided.

Demo:
https://imgur.com/a/qryItHH

Addresses:
#173


vidorteg:
> It works well! LGTM. Are we changing this to be the default option for the extension?

Not in this PR, but we can change that easily in the future :)

PR #195, opened 2020-08-24
[Scripts]: Adding script to extract from full edge zip

mliao95:
This helps improve workflow for targeting different versions.  How the script works is that given a full Edge zip, the script will unzip the necessary files to the `vscode-edge-devtools/out/edge/src` directory.

This PR also changes the `Contributing.md` to use the `extractEdgeFromFullZip.js` script for targeting different versions instead of the legacy steps.  Legacy method use case now limited to only if the scripts break.


mliao95:
Converting to draft since this only works for Windows and we should add Mac support before setting up the PR.

PR #197, opened 2020-08-25
[Changelog]: Updating version to 84.0.522.63

mliao95:


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 1.0.10
```
vidorteg, line 1: I think is a good idea to move to the increase the minor (major.minor.revision) every time we update the Devtools version (to a new major)

mliao95: Sounds good.  I don't think we have been doing that before, but we might as well start now.

vidorteg (approved):
Tested and it is working.

PR #200, opened 2020-09-02
Upgrades dependencies to latest stable version

vidorteg:
This PR updates all dependencies to their latest stable versions. It also contains a couple of fixes that were created as part of the migration mostly for ts-lint and updating the typescript version.


mliao95 (approved):
Tested and LGTM!

PR #201, opened 2020-09-18
[Contributing]: Removing environment variable from setup process

mliao95:
Since we are storing the Edge files within the vscode-edge-devtools directory, we no longer need environment variables and can use relative pathing.  This PR removes the need for `EDGE_CHROMIUM_PATH` and `EDGE_CHROMIUM_OUT_DIR` environment variables.

### DIFF
--- a/postBuildStep.ts
+++ b/postBuildStep.ts
@@ -39,7 +39,7 @@ async function copyStaticFiles() {
     await fse.ensureDir(commonOutDir);
     await copyFile(commonSrcDir, commonOutDir, "styles.css");
 
-    const dirName = __dirname.replace('\\', '/');
+    const dirName = __dirname.replace(/\\/g, '/');
     const sourceFilesPath = dirName + '/out/edge/src';
 
     const toolsSrcDir =
--- a/postBuildStep.ts
+++ b/postBuildStep.ts
@@ -39,25 +39,21 @@ async function copyStaticFiles() {
     await fse.ensureDir(commonOutDir);
     await copyFile(commonSrcDir, commonOutDir, "styles.css");
 
-    const dirName = __dirname.replace(/\\/g, '/');
-    const sourceFilesPath = dirName + '/out/edge/src';
+    const sourceFilesPath = path.normalize(__dirname + '/out/edge/src');
 
-    const toolsSrcDir =
-        `${sourceFilesPath}/third_party/devtools-frontend/src/front_end/`;
+    const toolsSrcDir = path.normalize(`${sourceFilesPath}/third_party/devtools-frontend/src/front_end/`);
     if (!isDirectory(toolsSrcDir)) {
         throw new Error(`Could not find Microsoft Edge DevTools path at '${toolsSrcDir}'. ` +
             "Did you run the 'npm run download-edge' script?");
     }
 
-    const toolsGenDir =
-        `${sourceFilesPath}/out/Release/gen/devtools/`;
+    const toolsGenDir = path.normalize(`${sourceFilesPath}/out/Release/gen/devtools/`);
     if (!isDirectory(toolsGenDir)) {
         throw new Error(`Could not find Microsoft Edge output path at '${toolsGenDir}'. ` +
             "Did you run the 'npm run download-edge' script?");
     }
 
-    const toolsResDir =
-        `${sourceFilesPath}/out/Release/resources/inspector/`;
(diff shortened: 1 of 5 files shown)


postBuildStep.ts
```diff
@@ -38,27 +38,25 @@ async function copyStaticFiles() {
     await fse.ensureDir(commonOutDir);
     await copyFile(commonSrcDir, commonOutDir, "styles.css");
 
-    // Must set environment variables EDGE_CHROMIUM_PATH and EDGE_CHROMIUM_OUT_DIR
-    // E.g. set EDGE_CHROMIUM_PATH=F:/git/Edge/src
-    //      set EDGE_CHROMIUM_OUT_DIR=Release
-    // See CONTRIBUTING.md for more details
+    const dirName = __dirname.replace(/\\/g, '/');
+    const sourceFilesPath = dirName + '/out/edge/src';
```
vidorteg, line 42: lets use path to normalize dirName and to join dirName with out/edge/src. That way we don't neet to worry about \ vs / on linux, mac or windows.

scripts/downloadAndExtractEdge.js
```diff
@@ -69,16 +69,14 @@ async function downloadZipFile(downloadUrl) {
       fs.unlink('edge.zip', () => {} );
       let dirName = __dirname;
       if (isWindows) {
-        flipSlashDirName = dirName.replace(/\//g, '\\');
+        const flipSlashDirName = dirName.replace(/\//g, '\\');
```
vidorteg, line 72: same here

src/test/helpers.ts
```diff
@@ -139,15 +139,29 @@ export function getFirstCallback(mock: jest.Mock, callbackArgIndex: number = 0):
 
 /**
  * Returns the contents of the specified file, if the file is not found returns null
- * @param uri The uri relative to the 'gen' folder specified in EDGE_CHROMIUM_PATH environment variable.
+ * @param uri The uri relative to the 'gen' folder.
  */
 export function getTextFromFile(uri: string) {
+    // Changing from C:\[vscode_directory]\src\test to C:/[vscode_directory]
+    const dirName = removeLastTwoDirectories(__dirname.replace(/\\/g, '/'));
+    const sourceFilesPath = dirName + '/out/edge/src';
```
vidorteg, line 147: same here

src/test/helpers.ts
```diff
@@ -139,15 +139,29 @@ export function getFirstCallback(mock: jest.Mock, callbackArgIndex: number = 0):
...
     }
 
     return null;
 }
+
+/**
+ * @param filepath
+ */
+function removeLastTwoDirectories(filepath: string) {
+    const arr = filepath.split('/');
```
vidorteg, line 163: instead of doing arr.pop() twice, we could use path and select the specific part we are interested in.

vidorteg (changes requested):
Left some changes, most of them related to the use of path library to handle paths instead of doing it manually.

src/test/helpers.ts
mliao95: Looking at the path library, I could not find a function that easily parsed out and selected which directory I wanted.  I improved the code to:
```
    const arr = filepath.split(path.sep);
    arr.splice(-2);
    return arr.join(path.sep);
```

I think this is as simple as I can get it.

src/test/helpers.ts
mliao95: Fixed.

scripts/downloadAndExtractEdge.js
mliao95: `downloadAndExtractEdge.js` is a standalone isolated script, so we cannot import the path library unless we list this as a module.  Don't know if we want to go down that route especially if we are already version checking for which Edge source files to download.

postBuildStep.ts
mliao95: Fixed.  Didn't realize that the path library was perfect for this kind of thing.  Thanks!

scripts/downloadAndExtractEdge.js
vidorteg: you are right we are already checking for all the platforms :)

src/test/helpers.ts
vidorteg: this is good! thanks Michael. I would post the function I was talking about but I spent 15 minutes looking for it and could not find it... I'll take note if I see it again 👍

vidorteg (approved):

PR #219, opened 2020-10-13
Re-disable quick open menu and filter command menu options

mliao95:
Quick open and command menu should be disabled via this commit: https://github.com/microsoft/vscode-edge-devtools/pull/131

This PR re-disables the Quick open menu, but leaves the command menu available with filtered commands.  The commands will contain the commands from the available tabs (Only Elements commands when Network is disabled and Elements and Network commands when Network is enabled.)

command-menu


vidorteg (approved):
LGTM

PR #221, opened 2020-10-16
Fixing ${workspaceFolder} when overriding a map.

vidorteg:
**Summary**
Absolute path mapping was working but using `${workspaceFolder}` placeholder was not.

**Root:**
There were different issues causing the same problem:
- The API we were using to cast {workspace} stopped working.
- Test should have caught this, but the tests were mocking this API.
- There was a missing mock needed for testing the actual execution path( fse.pathExists.)

**Fix:**
- Using `${workspaceFolder}` when overriding a path works correctly now.
- Changed the tests to validate the following 3 scenarios:
  - Get the default path mapping
  - Override path mapping with ${workspaceFolder} placeholder.
  - Override  mapping with an absolute path.


src/utils.ts
```diff
@@ -427,10 +437,7 @@ export function applyPathMapping(
         const wildcardValue = overridePatternMatches[1];
         let mappedPath = rightPattern.replace(/\*/g, wildcardValue);
         mappedPath = debugCore.utils.properJoin(mappedPath); // Fix any ..'s
-        mappedPath = mappedPath.replace(
-            "${workspaceFolder}",
-            vscode.workspace.workspaceFolders ? vscode.workspace.workspaceFolders[0].uri.toString() : "" || "");
-
+        mappedPath = replaceWorkSpaceFolderPlaceholder(mappedPath);
```
mliao95, line 440: There is a new behavior here after the change.  If `vscode.workspace.workspaceFolders` is `null` or `undefined`, in this branch, the returned path will still contain `${workspaceFolder}` in the path.  Before this change, the mapped path would remove the `${workspaceFolder}`.

Is this intentional?

src/utils.ts
```diff
@@ -446,6 +453,21 @@ function isHeadlessEnabled() {
     return headless;
 }
 
+/**
+ * Replaces the workspaceFolder placeholder in a specified path, returns the
+ * given path with file disk path.
+ * @param customPath The path that will be replaced.
+ */
+function replaceWorkSpaceFolderPlaceholder(customPath: string) {
+    if (vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders[0].uri.toString()) {
```
mliao95, line 462: small NIT here:
might be worth assigning `vscode.workspace.workspaceFolders[0]` or `vscode.workspace.workspaceFolders[0].uri.toString()` to a `const` for readability's sake.  This might be confusing since we are specifically using the `0` index of the `workspaceFolders` array without an explanation for what it is.

This can be resolved through a comment as well if you prefer that method.

mliao95 (commented):
Generally LGTM with two comments.

src/utils.ts
vidorteg: Yes, this is part of the  bug fix. In this specific case you have:
`await transformer.fixSource(localSource)` a couple of lines below which should take care of correctly mapping sourcePath. I also enabled a test for this scenario (turned sourceMaps true).

mliao95 (approved):
LGTM

PR #223, opened 2020-10-20
Implementing listeners for mac meta key

mliao95:
Mac meta key (command key) is not propagated out of the DevTools IFrame.  This change creates a keydown listener inside the iframe to execute shortcut commands using the meta key.

Functionalities include:
- copy ( meta + C )
- cut ( meta + X )
- paste ( meta + V )
- select all ( meta + A )
- undo ( meta + Z )
- redo ( meta + Y )


codepo8:
Works on my machine now, This is great.

codepo8 (approved):
Works here and looks like a sensible set of presets. We can extend if people want more.

vidorteg (approved):

PR #233, opened 2020-10-29
Including Debugger for Microsoft Edge as a dependency

vidorteg:
Some users assume that debugging capabilities are part of the Devtools extension, so making that the default behavior to avoid confussion.


mliao95 (approved):
LGTM

PR #235, opened 2020-11-04
[Update]: Upgrading Edge DevTools version to 85.0.564.40

mliao95:
Upgrading to version 85.0.564.40.

Changes:
- Adding `applyRemovePreferencePatch` and `applyDefaultTabPatch`
    - `applyRemovePreferencePatch` patches the `removePreference` function such that it does not try to delete objects from `localStorage` which is set to `undefined` in the extension
    -`applyDefaultTabPatch` removes the need of `TabbedLocation._defaultTab` which was causing issues with our `applyAppendTabPatch`.
- Removing `applyShowElementsTab`
    - removing this patch that sets `_defaultTab` to `elements`
    - This is generally unnecessary as it will default to Elements if Network is disabled and we do not want to pigeonhole users to the Elements panel if Network is enabled.
- Updating extension version to `1.1.2`


vidorteg (approved):
Tested, works as expected.

PR #239, opened 2020-11-24
[Polyfill]: Improve extension settings infrastructure

mliao95:
This PR standardizes and improves the workflow of grabbing extension settings values to use within the DevTools.  We create a global map `Root.Runtime.extensionSettings` that we fill with values on the constructor of Runtime.  Values such as `Root.Runtime.extensionSettings.get('networkEnabled')` or `Root.Runtime.extensionSettings.get('theme')` can be used throughout the DevTools without needing to utilize the InspectorFrontendHost callback each time we need to access these settings.


src/devtoolsPanel.test.ts
```diff
@@ -467,18 +467,18 @@ describe("devtoolsPanel", () => {
                 expect(mockVsCode.window.showTextDocument).toHaveBeenCalled();
             });
 
-            it("calls getApprovedTabs", async () => {
+            it("calls getNetworkSetting", async () => {
```
vidorteg, line 470: why is approvedTabs getting renamed to networkSettings, my understanding is that one will retrieve the list of tabs that are allowed to be displayed while the other one seems to retrieve specific settings for network.

src/host/polyfills/simpleView.ts
```diff
@@ -29,8 +29,65 @@ export function revealInVSCode(revealable: IRevealable | undefined, omitFocus: b
     return Promise.resolve();
 }
 
-export function getApprovedTabs(callback: () => void) {
-    InspectorFrontendHost.InspectorFrontendHostInstance.getApprovedTabs(callback);
+export function getNetworkSetting(callback: () => void) {
+    // @ts-ignore we inject this function into the Runtime constructor where we call InspectorFrontendHost, not InspectorFrontendHost.InspectorFrontendHostInstance
```
vidorteg, line 33: currently InspectorFrontendHost.InspectorFrontendHostInstance seems to be correctly typed. Why is the ts-ignore required in here?

vidorteg (changes requested):
Left a couple of comments to avoid blocking. I still need to  take a deeper review to this one as there are still some open questions on my side.

src/host/polyfills/simpleView.ts
mliao95: Fixed

src/devtoolsPanel.test.ts
mliao95: Talked offline:
- creating new pattern of `get{SETTINGS_NAME}Setting` getter functions for functions that check the vscode settings for the extension
- Landed on `getNetworkSetting`.  The `getApprovedTabs` function would check whether the network pane was enabled/disabled in the vscode settings, so the `getNetworkSetting` name makes sense.

vidorteg (approved):
LGTM

PR #244, opened 2020-12-09
[Keyboard Shortcuts]: Adding focus next/previous editor listeners inside dtwindow to quickly switch between panes

### LINKED ISSUE #208: Switching between tools and editor with keyboard shortcut
Is there a fast way to navigate between the Edge DevTools extension and an open document in the editor? I like to use the keyboard (arrow keys) to navigate through the DOM in the Edge DevTools. Is there a keyboard shortcut that switches between the editor and the elements panel? Or a way to get to the VS code Command Palette when using keyboard navigation inside the Elements panel? Let's say the focus is on the H1 element in the devtools:

image

The way I can go from there to my index.html is: shift + tab, tab, tab, tab. ctrl + page down. It would be nice if there is a faster way.

mliao95:
The DevTools window traps keyboard shortcuts inside the iframe, so users who are focused inside the DevTools cannot navigate in/out using `CTRL + PageDown` or `CTRL + PageUp`.  This change implements the next/previous editor shortcut keys within the DevTools window to allow users to easily switch between windows with the next/previous shortcuts.

closes #208

### DIFF
--- a/src/common/webviewEvents.ts
+++ b/src/common/webviewEvents.ts
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 
 export type WebviewEvent = "getState" | "getUrl" | "openInEditor" | "ready" | "setState" | "telemetry" | "websocket"
-| "getApprovedTabs" | "getThemes" | "copyText";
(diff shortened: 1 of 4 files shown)


vidorteg (commented):
LGTM

vidorteg (approved):
LGTM

PR #248, opened 2021-01-05
Add "single close" buttons to instances panel 

### LINKED ISSUE #186: [headless instances] There should be a way to delete the instances in the list of targets
As we aren't opening a new window with the headless instances there is no way for the user to close the sessions. A sensible way would be to add a close icon next to each instance to be able to close it from VS Code.

vidorteg:
Fixes #186 

Users wanted a way to close individual instances on the target list, this is specially important when headless mode is active. So this PR fixes this issue.


mliao95 (approved):
LGTM

Issue #259, opened 2021-01-24
"Jump to file" could behave more like the Edge developer tools in the browser


billie-spektor:
When clicking on a style property in the styles pane of the extension, you jump to the corresponding CSS file. This is very nice, but I would suggest to align this click behaviour with the behaviour in the Edge devtools in the browser:

- when clicking on a style property, don't jump to the file
- when clicking on the filename, do jump to the file
- when control-clicking on a selector, style property or value, do jump to the corresponding file and line

The reason for this suggestion is that I sometimes just want to alter a value and do some minor CSS tweaking on a related component, while working in the context of another component. I don't want to switch context to the related component, because I'm just temporarily trying some things out to see how it interacts with the component I'm working on. Therefore, I think it could be nice if the file-jumping is a bit more of an explicit user action (by only jumping to the file when the user clicks on the filename, or when CTRL is used).

codepo8:
This should work when you have your files synced. As we don't have the sources tab like we have in the developer tools, I wonder where the link should go. I agree though that it shouldn't be a link if it doesn't go anywhere.

mliao95:
Hi billie-spektor.  After testing around, I can see your frustration with this behavior.  I'll take a look and try implementing a better workflow for this.  Thanks for filing this issue!

billie-spektor:
Thanks :) The gif in the pull request looks good!

PR #264, opened 2021-01-29
[VSCode Revealer]: Fix revealer workflow for style properties sections

mliao95:
This PR changes revealing workflow for clicking style properties and style sheet links.

Issues:
- VSCode reveals style sheet whenever a style property is clicked
- Clicking on a style sheet in the styles properties section does not reveal anything

Changes:
- VSCode only reveals the style sheet when the user `CTRL + Click` on the style property
- Clicking on a style sheet link in the styles properties section will reveal the style sheet in VSCode.

GIF:
extension-fix-revealer

### DIFF
--- a/src/host/polyfills/simpleView.ts
+++ b/src/host/polyfills/simpleView.ts
@@ -91,11 +91,8 @@ export function applyCommonRevealerPatch(content: string) {
 
 export function applyStylesRevealerPatch(content: string) {
     const pattern = /this\._navigateToSource\(selectElement,\s*true\);/g;
-    if (content.match(pattern)) {
-        return content.replace(pattern, '');
-    } else {
-        return null;
-    }
+    const replacementText = '';
+    return replaceInSourceCode(content, pattern, replacementText);
 }
 
 export function applyQuickOpenPatch(content: string) {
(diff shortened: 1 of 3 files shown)


src/host/polyfills/simpleView.ts
```diff
@@ -92,6 +92,24 @@ export function applyCommonRevealerPatch(content: string) {
...
+    const pattern = /this\._navigateToSource\(selectElement,\s*true\);/g;
+    if (content.match(pattern)) {
+        return content.replace(pattern, '');
+    } else {
+        return null;
+    }
+}
+
+export function applyLinkActionsPatch(content: string) {
+    const pattern = /\"Elements panel\"/g;
```
vidorteg, line 105: we should be more specific with this CL, I can see a lot of hits.

vidorteg (approved):
LGTM

src/host/polyfills/simpleView.ts
mliao95: Removed this function - not needed for most recent iteration.

vidorteg (approved):

billie-spektor:
Hi mliao95, I just noticed that clicking on the property (without ctrl) does bring me to the CSS file. Is this intentional behaviour or regression?

mliao95:
Hey @billie-spektor , thanks for bringing this up.  This is an intentional feature. Is this interrupting your workflow? If so, we'd love to hear about it and see what we can do from our side to provide a better experience.

billie-spektor:
Hi @mliao95,

We previously talked about the reasons in #259; it mainly has to do with avoiding unintended context switching. When I'm working on a CSS component in file 1, I might want to quickly tweak or toggle a property for a different component in file 2 to see how these components interact. 

Let's say component A is an animated button that moves in to the page from the bottom. This component is finished and now I'm working on component B, which is a picture that is on the page and in part behind the animated button. I would like to be able to easily alter the position of the button, maybe by temporarily turning off the animation, so that it doesn't interfere with the styling of component B, the picture. When I toggle the animation property, I'm (unintentionally) taken to the file for component A, even though that component is already finished. The only reason to turn the animation off was to avoid noise when styling component B.

I do like the ability to directly go to the file, but for me it works best if it's an explicit choice and not default behavior that is always triggered when experimenting and tweaking properties. Requiring control click (instead of just clicking) to jump to a file location also would be consistent with the implementation in the Edge developer tools in the browser. 

If you prefer the current behavior, please consider a configuration option to alter this behavior.

PR #268, opened 2021-02-02
[SimpleView]: Create and use helper function to replace code

### LINKED ISSUE #266: [SimpleView]: Create helper function to reduce repeated code
A ton of the code in `simpleView.ts` follow the same pattern of:

```
    const pattern = {Regex};
    const match = content.match(pattern);
    if (match) {
        return content.replace(pattern, {replacementString});
    } else {
        return null;
    }
 ```
 
We can greatly cut down on the code and improve readability by creating a helper function that mimics this behavior.

### PR DESCRIPTION
**Pull request by mliao95**
Creating helper function called `replaceInSourceCode` in `simpleView.ts` to help with finding the injection point and replacing code.

```
function replaceInSourceCode(content: string, pattern: RegExp, replacementText: string, keepMatchedText?: KeepMatchedText) {
    const match = content.match(pattern);
    if (match) {
        if (keepMatchedText) {
            const matchedText = match[0];
            if (keepMatchedText === KeepMatchedText.AtEnd) {
                replacementText = `${replacementText}${matchedText}`;
            } else {
                replacementText = `${matchedText}${replacementText}`;
            }
        }
        return content.replace(pattern, replacementText);
    } else {
        return null;
    }
}
```

This change helps cut down on almost 100 lines of code and simplifies adding a new patch.

Additionally, I have split apart any patches that contained two separate code injections so that we isolate any patching issues to one specific injection site.

closes #266

### DIFF
--- a/postBuildStep.ts
+++ b/postBuildStep.ts
@@ -7,14 +7,17 @@ import applyPaddingInlineCssPatch from "./src/host/polyfills/cssPaddingInline";
 import { applyContentSecurityPolicyPatch } from "./src/host/polyfills/inspectorContentPolicy";
 import applyRuntimeImportScriptPathPrefixPatch from "./src/host/polyfills/runtime";
 import {
-    applyAppendTabPatch,
+    applyAppendTabOverridePatch,
+    applyAppendTabConditionsPatch,
     applyCommonRevealerPatch,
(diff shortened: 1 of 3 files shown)


vidorteg (approved):
Thanks for tackling all of these Michael. LGTM

PR #275, opened 2021-02-05
[Feature]: Rerouting DevTools console messages to VSCode output window

### LINKED ISSUE #126: how do I view the console output?
not sure if it's supported yet or not, but I couldn't find anywhere that fetch the console outputs into the elements

mliao95:
Issue:
- Console messages cannot be read in the DevTools extension

Changes:
- Add function that sends all console messages to the VSCode output window

GIF:
extension-console-output

closes #126


src/host/toolsHost.ts
```diff
@@ -13,6 +13,7 @@ import ToolsResourceLoader from "./toolsResourceLoader";
 import ToolsWebSocket from "./toolsWebSocket";
 
 export default class ToolsHost {
+    [x: string]: any;
```
vidorteg, line 16: I don't think this is needed.

vidorteg (approved):
LGTM - 1 minor comment

src/host/toolsHost.ts
```diff
@@ -16,6 +16,8 @@ export default class ToolsHost {
     private resourceLoader: Readonly<ToolsResourceLoader> | undefined;
     private getHostCallbacksNextId: number = 0;
     private getHostCallbacks: Map<number, (preferences: object) => void> = new Map();
+    // We need to add a dummy property to get around build errors for sendToVscodeOutput.
+    InspectorFrontendHostInstance: any;
```
mliao95, line 20: @vidorteg 

This does the same thing as the original `[x: string]: any;` which is to make it so this function doesn't throw an error since `InspectorFrontendHost.InspectorFrontendHostInstance` isn't defined:
```
export function sendToVscodeOutput(message: string) {
    InspectorFrontendHost.InspectorFrontendHostInstance.sendToVscodeOutput(message);
}
```

The main issue is that if we want to reference InspectorFrontendHost in `root.js` like in `getVscodeSettings()`, we use `InspectorFrontendHost`.  However, in any file outside of `root.js` like for `sendToVscodeOutput`, we need to call `InspectorFrontendHost.InspectorFrontendHostInstance`.

We're kind of using these functions unconventionally since we declare them in `simpleView.ts` but we never actually use them directly and only use it to inject the function in our source code.  So even though it doesn't need to make sense in the context of our extension code, it is giving us an error.

I'll wait to check this change in until you see this.  Let me know if there is a better solution than creating this dummy property.

vidorteg (approved):
LGTM

PR #279, opened 2021-02-09
[Screencast]: Change cursor type from touch to default

### LINKED ISSUE #104: Allow modifiying cursor style
When using "Elements" tab the cursor transforms to a dot which is not very precise.
Moreover, when debugging desktop versions you may want to check the cursor styles when hovering buttons or any other action that changes cursor style.

It should be an option to use default cursor instead of this dot cursor.

mliao95:
Issue:
- Screencast uses a touch cursor, but we aren't targeting mobile devices in the extension

Changes:
- Change cursor to default type for screencast view

extension-default-cursor

closes #104

### DIFF
--- a/postBuildStep.ts
+++ b/postBuildStep.ts
@@ -31,6 +31,7 @@ import {
     applyRemoveBreakOnContextMenuItem,
     applyContextMenuRevealOption,
     applyRemovePreferencePatch,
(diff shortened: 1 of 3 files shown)


vidorteg (approved):
LGTM

PR #284, opened 2021-02-16
[Settings]: Create setting to show service and shared workers in the target list (will now hide by default)

### LINKED ISSUE #276: Filter extensions and service workers from list of targets
In the list of targets, I can connect to many targets, even though I only have one tab open in Edge. I'm not sure if there is a use  case for connecting to chrome-extensions and service workers. If not, I think it would be a good idea to limit the list of targets to actual tabs.

mliao95:
This PR hides service and shared workers behind a settings flag to declutter the target list.

GIF:
settings-show-workers

closes #276

### DIFF
--- a/src/cdpTargetsProvider.ts
+++ b/src/cdpTargetsProvider.ts
@@ -6,7 +6,7 @@ import TelemetryReporter from "vscode-extension-telemetry";
 import * as path from "path";
 import * as fs from "fs";
 import CDPTarget from "./cdpTarget";
-import { fixRemoteWebSocket, getListOfTargets, getRemoteEndpointSettings, IRemoteTargetJson } from "./utils";
+import { fixRemoteWebSocket, getListOfTargets, getRemoteEndpointSettings, IRemoteTargetJson, SETTINGS_STORE_NAME } from "./utils";
 
 export default class CDPTargetsProvider implements vscode.TreeDataProvider<CDPTarget> {
     public readonly onDidChangeTreeData: vscode.Event<CDPTarget | null>;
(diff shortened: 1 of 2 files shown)


src/cdpTargetsProvider.ts
```diff
@@ -43,9 +43,7 @@ export default class CDPTargetsProvider implements vscode.TreeDataProvider<CDPTa
                         let targetsProcessed = 0;
                         responseArray.forEach(async (target: IRemoteTargetJson) => {
                             const actualTarget = fixRemoteWebSocket(hostname, port, target);
-                            if (actualTarget.type !== 'page') {
-                                targets.push(new CDPTarget(actualTarget, "", this.extensionPath));
-                            } else {
+                            if (actualTarget.type !== 'service_worker' && actualTarget.type !== 'shared_worker') {
```
vidorteg, line 46: Is there a scenario where we want to use the extension for service worker debugging? Maybe with the network tab? If this is the case I think a setting like like:
"Show service workers on target list" and make it false by default makes more sense. Thoughts?

vidorteg (changes requested):
Left a comment

src/cdpTargetsProvider.ts
mliao95: I like that idea just so we don't break any existing workflows.  Will work on a fix now.

mliao95: Fixed.  See new GIF in description

vidorteg (approved):
Nice :)

PR #287, opened 2021-03-02
[Bug]: fix *.launch *.refresh not found bug

### LINKED ISSUE #277: Error: "command 'vscode-edge-devtools-view.launch' not found
Seeing this error pop up occasionally when trying to create a new target.  In my case, this is a transient issue where trying to launch again generally fixes it.

mliao95:
Issue:
- This message pops up when users press either the `Open a new tab` or `refresh` buttons at the top before the commands can be registered
image

Changes:
- added `titleCommandsRegistered` condition to show the `Open a new tab` or `refresh` buttons
    - the `titleCommandsRegistered` flag is turned to true once registration for those commands are complete.

closes #277

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -98,6 +98,7 @@ export function activate(context: vscode.ExtensionContext) {
     context.subscriptions.push(vscode.commands.registerCommand(
         `${SETTINGS_VIEW_NAME}.copyItem`,
         (target: CDPTarget) => vscode.env.clipboard.writeText(target.tooltip)));
+    vscode.commands.executeCommand('setContext', 'titleCommandsRegistered', true);
 }
 
 export async function attach(
--- a/package.json
+++ b/package.json
@@ -435,12 +435,12 @@
             "view/title": [
                 {
                     "command": "vscode-edge-devtools-view.launch",
-                    "when": "view == vscode-edge-devtools-view.targets",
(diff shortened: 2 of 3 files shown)


src/extension.ts
```diff
@@ -1,234 +1,235 @@
...
-            const match = exeName!.match(/(chrome|edge)/gi) || [];
-            const knownBrowser = match.length > 0 ? match[0] : "other";
-            const browserProps = { exe: `${knownBrowser.toLowerCase()}` };
-            telemetryReporter.sendTelemetryEvent("command/launch/browser", browserProps);
-        }
-        browserInstance = await launchBrowser(browserPath, port, url, userDataDir);
-        await attach(context, url, config);
-    }
-}
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
vidorteg, line 1: not sure why but the file shows all the content as new/changed which makes reviewing it really difficult. Could you double check and scope the review to the changes?

vidorteg (changes requested):
Left some comments

src/extension.ts
mliao95: It looks like this is happening whenever we use Github's rebase... Not sure why, but I'll try fixing this.

mliao95: Okay fixed and ready for review

vidorteg (approved):
This seems like a difficult issue to spot, thanks for fixing it @mliao95

PR #288, opened 2021-03-03
[Feature]: Add dropdown menu in title view containing link to settings and changelog

### LINKED ISSUE #255: Enable a changelog button on the target list
Enable a link on the "target list" panel that redirects you to the changelog post in the blog.

mliao95:
This PR adds a dropdown menu in the extension's title view that contains links to the extension's settings and changelog.

GIF:
extension-more-actions

closes #255

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,6 +97,16 @@
                 "command": "vscode-edge-devtools-view.copyItem",
                 "category": "Microsoft Edge Tools",
                 "title": "Copy Value"
+            },
+            {
+                "command": "vscode-edge-devtools-view.openSettings",
+                "category": "Microsoft Edge Tools",
+                "title": "Open Settings"
+            },
+            {
+                "command": "vscode-edge-devtools-view.viewChangelog",
+                "category": "Microsoft Edge Tools",
+                "title": "View Changelog"
             }
         ],
         "configuration": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -437,6 +446,11 @@
                     "command": "vscode-edge-devtools-view.refresh",
                     "when": "view == vscode-edge-devtools-view.targets",
                     "group": "navigation"
+                },
+                {
```
joselea, line 450: what is the difference between these 2 commands? the registration that has a title and icon and the second one that has a when and a group?

joselea (commented):
LGTM, have a quick question

package.json
mliao95: The first section with the title and icon is the actual registration of the command whereas the second one is to place the existing command to the extension title bar like this:
image

joselea (approved):
LGTM

vidorteg (changes requested):
I like the change but I'm not sure if the toolbar is the best place for it, as it feels out of place with the other buttons (which are extension commands). I took a look to other extensions and some of them add a "..." menu at the end which triggers a context menu where we could add the changelog (along with maybe other options like "Settings"). similar to this:
image

vidorteg (approved):
Looks incredible @mliao95 !, thanks for the quick turnaround

PR #300, opened 2021-03-17
Side panel auto-refreshes when targets are created or destroyed

flynnolivia:
Changes in this PR:
- Add event listeners to the browser instance once it's obtained, so we can refresh the side panel when target creation/destruction events are emitted
- Ensure tests still pass by mocking the browser instance's addListener function

### DIFF
--- a/src/extension.test.ts
+++ b/src/extension.test.ts
@@ -363,6 +363,7 @@ describe("extension", () => {
     });
 
     describe("launch", () => {
+        const fakeBrowser = {addListener: () => null};
         let mockReporter: Mocked<Readonly<TelemetryReporter>>;
         let mockUtils: Partial<Mocked<typeof import("./utils")>>;
         let mockPanel: Partial<Mocked<typeof import("./devtoolsPanel")>>;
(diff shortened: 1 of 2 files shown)


vidorteg (changes requested):
I tested the change but could not get the expected behavior. This is the scenario I tested:
- Open the extension (non-headless)
- Open a new browser target
- (On the browser) Open a new tab
- Go to contoso.com. <-----------At this point the target should be visible on the list
- Go to extension (assuming all targets are visible)
- Close the contoso tab <------ At this point the target list should not have contoso listed.

Maybe I'm not understanding correctly the bug?

flynnolivia:
I think this scenario should be sufficient to test out the bug. I recorded this gif of the behavior on my local branch, can you take a look and let me know if this is the expected behavior? https://imgur.com/a/kcZn6ed

vidorteg:
it is working as expected, it was probably a configuration on my side. Thanks for double checking and the fix @flynnolivia

vidorteg (approved):

PR #302, opened 2021-03-20
[Version Update]: Upgrade Edge DevTools to 88.0.705.9

mliao95:
This PR updates the Edge DevTools to version 88.0.705.9 and modifies some patches to maintain compatibility.

Additionally, this PR patches the `Move to bottom/top` context menu option out of the extension DevTools.
image

This functionality is only relevant for the browser DevTools where there are many more tool options.  Furthermore, allowing this functionality will require more safeguarding patches to prevent the extension from entering a broken state.  IMO the minimal value this provides is not worth the needed safeguarding patches, so disabling altogether seems like the right approach.

Example:
- Disable network tool and what's new so drawer is disabled
- Move elements tool to bottom drawer via context menu
- Extension in a broken state where Elements tool is in the drawer, but the drawer is disabled


scripts/downloadAndExtractEdge.js
```diff
@@ -36,7 +37,7 @@ function fetchDownloadUrl(version) {
   }
   console.log(`Downloading Microsoft Edge DevTools version ${fullVersion}`);
   for (let object of jsonObjects) {
-    if (object.product === 'Microsoft Edge DevTools' && object.release === fullVersion && object.platform === 'Windows x64') {
```
hkal, line 39: Why was this check removed?

hkal (approved):
LGTM

scripts/downloadAndExtractEdge.js
mliao95: We stopped producing different source file downloads for different platforms since it's all the same.  For this latest version, I left the platform blank, so we need to remove this check.

vidorteg (approved):
everything LGTM. I agree no point on keeping the button in the extension. Thanks for the fix @mliao95

PR #314, opened 2021-03-23
Fixing download-edge script in ARM devices

vidorteg:
donwload-edge script fails on ARM devices on a race condition between extracting the edge.zip and deleting the file. This PR fixes the issue allowing it to finish the extract before deleting.

### DIFF
--- a/scripts/downloadAndExtractEdge.js
+++ b/scripts/downloadAndExtractEdge.js
@@ -56,11 +56,11 @@ async function downloadZipFile(downloadUrl) {
   await fs.remove('out/edge');
 
   const file = fs.createWriteStream('edge.zip');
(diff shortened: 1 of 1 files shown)


mliao95:
Do you happen to know why this is ARM devices specifically?  Also could you link the relevant issues this fixes?  Like will we need to do anything with default browser paths or was this the main issue behind ARM devices?

vidorteg:
I think is has to do with computing power.  It is probably taking more time to the unzipper to complete which fails the deletion (or viceversa), this is not related to the extension not working on the ARM devices, I have not gotten there yet as I got blocked by this one first. Here is a screenshot of the error:
image

mliao95 (approved):
Got it, thanks.  Makes sense. LGTM

PR #316, opened 2021-03-23
Fixing close-instance button flakiness

flynnolivia:
This PR adds a check to the close-instance function so that it returns early if the argument is undefined. The error came from the close-instance function being called with an undefined argument, so this seems to happen when the button is clicked after the target is destroyed, but before the view is finished refreshing to reflect that (I was able to repro consistently by spamming the close button).

This prevents the error from popping up, but the behavior might be confusing, since the target is still visible and the close-instance button can still be pressed for some time after initially pressing it (especially if it's pressed a lot+quickly, it has to refresh many times and that seems to cause it to hang). Maybe we could improve on this by disabling the close button after it's clicked once, but I'm not sure if there's a way to specify that in package.json or on the TreeItem object. Let me know if I should look into it more!


src/extension.ts
```diff
@@ -81,6 +81,9 @@ export function activate(context: vscode.ExtensionContext) {
     context.subscriptions.push(vscode.commands.registerCommand(
         `${SETTINGS_VIEW_NAME}.close-instance`,
         async (target: CDPTarget) => {
```
hkal, line 83: Setting `target` as an optional parameter would catch this bug at transpile time (assuming the flakiness was caused by an NFE on line 73).  

Were you able to root cause the scenario where `target` is null or undefined? We may need to make other functions that rely on targets to assume they're optional as well (i.e line 70).

flynnolivia: Ok, I'll make it an optional parameter (the flakiness is from NPE on line 90 though). I'm not exactly sure what the root cause of the undefined argument is yet, but the scenario is:
1. click the close button
2. the extension takes some time to process this, it shows the blue clock icon and animation at the top of the targets sidepane
3. click the close button while the extension is still processing

hkal: My bad, I typoed the error line. I meant line 90, where we access a property on the potentially null target.

flynnolivia: Line 70 seems to be the only other part similar to the one causing this issue. Do you think I should change that function to have an optional param too? I think that one's a lot less likely to cause an issue, but maybe it's an edge case

mliao95:
Thanks for the fix, Olivia!  Looks and works great.  

In regards to disabling the button, I had to do something similar with the "Open a new tab" and "Refresh" buttons on start up.  Take a look at my PR here where I set a condition in `package.json` and toggle it in `extension.ts`: https://github.com/microsoft/vscode-edge-devtools/pull/287/files

Unfortunately, it removes the button altogether instead of disabling and graying it out.  I remember it was tough finding the exact property to gray and disable a button, but I'm convinced it exists somewhere.  Doing some research now and will update if I find anything.  Adding a `when` condition should be a valid workaround in the meantime.

src/extension.ts
hkal: In regards to line 70, the scenario is the extension attaching to the target. If there's a case where an attach fails (I'm not sure what would cause this), we could get an undefined target.

Double check with Michael and Vidal since they know the extension lifecycle better than I do.

mliao95:
Okay figured it out.  We want to use an `enablement` property, but we need to add it to where the command is created, not where it is shown.  So for the example above, it would look something like this:

```
            {
                "command": "vscode-edge-devtools-view.launch",
                "category": "Microsoft Edge Tools",
                "title": "Open a new tab",
                "enablement": "titleCommandsRegistered",
                "icon": {
                    "light": "resources/light/launch.svg",
                    "dark": "resources/dark/launch.svg"
                }
            },
```

src/extension.ts
flynnolivia: Got it, @mliao95 what do you think?

mliao95: Correct me if I'm wrong, but the issue here is that we get an undefined target when we spam the `close-instance` button because the first call to `close-instance` destroys the target and before the first call can update the UI, the second+ calls are attempting to delete the already deleted target.

If that is the case, disabling the `close-instance` and `attach` buttons should prevent either command to be called until the first call to `close-instance` is finished and we can assume that the CDPTarget will not be undefined.  

If we think that there is another issue causing the `undefined` target, I would go with making both `attach` and `close-instance` take optional parameters and continue investigating.

src/extension.ts
```diff
@@ -80,14 +80,11 @@ export function activate(context: vscode.ExtensionContext) {
...
-        async (target: CDPTarget) => {
-            // skip if the target has already been deleted
-            if (!target)
-                return;
+        async (target?: CDPTarget) => {
 
             // update with the latest information, in case user has navigated to a different page via browser.
             cdpTargetsProvider.refresh();
-            const normalizedPath = new URL(target.description).toString();
+            const normalizedPath = target ? new URL(target!.description).toString() : "";
```
mliao95, line 87: I think we should keep the `!target` check and return early if true.  We would just be doing unnecessary work if we knew we had an undefined target.

src/extension.ts
```diff
@@ -80,11 +80,11 @@ export function activate(context: vscode.ExtensionContext) {
     }));
     context.subscriptions.push(vscode.commands.registerCommand(
         `${SETTINGS_VIEW_NAME}.close-instance`,
-        async (target: CDPTarget) => {
+        async (target?: CDPTarget) => {
```
mliao95, line 83: Let's also make the optional parameter and early return in `attach` as well.

src/extension.ts
```diff
@@ -81,10 +83,12 @@ export function activate(context: vscode.ExtensionContext) {
     context.subscriptions.push(vscode.commands.registerCommand(
         `${SETTINGS_VIEW_NAME}.close-instance`,
         async (target?: CDPTarget) => {
+            if (!target)
+                return;
 
             // update with the latest information, in case user has navigated to a different page via browser.
             cdpTargetsProvider.refresh();
-            const normalizedPath = target ? new URL(target!.description).toString() : "";
+            const normalizedPath = new URL(target!.description).toString();
```
mliao95, line 91: don't need the `!` anymore.

mliao95 (approved):
Thanks for the change and the investigation behind it!  LGTM.

PR #321, opened 2021-03-25
Replace TSLint with ESlint

vidorteg:
TSLint has been deprecated for some time now (~2y). This PR introduces ESlint although it DOES NOT enable the scanning yet as it throws hundreds of errors that needs to be fixed:

image

I think that a lot of the errors can be taken out based on consensus, but given that it has been years now, I prefer to proceed with the upgrade and force us to fix the issues.


mliao95 (approved):
LGTM

flynnolivia (approved):
LGTM too!

PR #325, opened 2021-03-26
Disable inline buttons when closing a target

flynnolivia:
Disables the buttons to prevent calling the close-instance function on a target that has already been deleted. This PR also replaces the temporary fix (have an optional parameter and return early if it's undefined).

Before:
disable-before
After:
disable-after

Alternatively, we could disable the close button and just not render the attach button once close is clicked-- it would reduce a little bit of unnecessary work, but I think the other way looks better. Let me know what you think!
disable-after2


mliao95 (approved):
Tested and LGTM

PR #327, opened 2021-03-26
[Bug Fix]: Restore screencast scroll functionality

### LINKED ISSUE #326: Screencast does not repaint after mouse events like scroll

mliao95:
In the update to version 88, a conditional was added for the `highlightInOverlay` call in the [DevTools source code](https://source.chromium.org/chromium/_/chromium/devtools/devtools-frontend/+/543c28e9b344ea2eee940e67ce707be0de1a8c07:front_end/screencast/ScreencastView.ts;l=239;drf=front_end%2Fscreencast%2FScreencastView.js;drc=a86b1a80e5a58a0e5d49c1fb5bdc69af5c67daca;bpv=1;bpt=0).  

```
      if (this._highlightNode) {
        this.highlightInOverlay({node: this._highlightNode, selectorList: undefined}, this._highlightConfig);
      }
```
In version 87, `highlightInOverlay` would be called with `this._highlightNode` being `undefined` when triggered via a mouse event such as scroll.  The addition of this conditional means that `highlightInOverlay` isn't triggered by scrolling or other non click mouse events.

It is actually safe to call `highlightInOverlay` in this scenario.  The purpose behind this added check was a change from an optional to required node parameter in `SDK.OverlayModel.HighlightData` which is the first param of `highlightInOverlay`.  

This PR replaces the `this._highlightNode` check with `true` such that it will repaint as it did in version 87

closes #326


vidorteg (approved):
LGTM

flynnolivia (approved):
LGTM

PR #332, opened 2021-03-30
[Accessibility]: Retain focus on property after toggling with context menu

mliao95:
This accessibility issue only repros in the extension and not in the browser version of the DevTools.  It is most likely due to the different focus behavior of context menus inside webviews.  When the context menu appears, it removes focus from the edited property whereas in the browser DevTools, the focus remains.  This patch checks for the property index before we create the context menu, so we can find and continue editing the property after toggling.

Before: 
property-focus-broken

After:
property-focus-fixed

Code:

Before:
```
    _handleContextMenuEvent(context, event) {
        const contextMenu = new ContextMenu.ContextMenu(event);
        if (this.property.parsedOk && this.section() && this.parent.root) {
            contextMenu.defaultSection().appendCheckboxItem(ls `Toggle property and continue editing`, async () => {
                const sectionIndex = this._parentPane.focusedSectionIndex();
                ...
                this._parentPane.continueEditingElement(sectionIndex, propertyIndex);
                ...
            }, !this.property.disabled);
        }
        ...
        contextMenu.show();
    }
```

After:
```
    _handleContextMenuEvent(context, event) {
        const contextMenu = new ContextMenu.ContextMenu(event);
        if (this.property.parsedOk && this.section() && this.parent.root) {
            const sectionIndex = this._parentPane.focusedSectionIndex();
            contextMenu.defaultSection().appendCheckboxItem(ls `Toggle property and continue editing`, async () => {
                ...
                this._parentPane.continueEditingElement(sectionIndex, propertyIndex);
                ...
            }, !this.property.disabled);
        }
        ...
        contextMenu.show();
    }
```


flynnolivia (approved):
Makes sense and the fix works on my end, lgtm!

mliao95:
> Makes sense and the fix works on my end, lgtm!
> I have a question though: is there any quick way to tell if a bug is from devtools being in a webview vs. the version not being up to date?

The best way is to check out the browser version of the DevTools (which is 88.0.705.9) and see if it worked there.  It ended up working there and not in the extension version even though they contain the same source code, so it must have been due to the environment.  We host the DevTools in an iframe inside a webview for the extension which has caused issues in the past, so that generally is the main culprit behind the discrepancies between same versions of the extension and the browser DevTools

vidorteg (approved):
LGTM, nice one 👍

PR #333, opened 2021-03-30
[Accessibility]: Announce success message after refreshing sidepane completes

flynnolivia:
In this PR:
- Trigger a notification at the end of `CDPTargetProviders.getChildren()` and hide it quickly to make the screen reader announce that the targets list has refreshed
- Updated tests

I'm not sure this is an ideal solution since it will also hide recent notifications that aren't related to our extension, but it seems to be what's available to fix the bug for now. There is a discussion at microsoft/vscode#114718 about exposing more API to control the screenreader which I can keep following.

### DIFF
--- a/src/cdpTargetsProvider.ts
+++ b/src/cdpTargetsProvider.ts
@@ -78,6 +78,9 @@ export default class CDPTargetsProvider implements vscode.TreeDataProvider<CDPTa
                     return a.targetJson.type < b.targetJson.type ? -1 : 1;
                 }
             });
+            // Raise and quickly hide a notification so the screen reader announces the new state of the targets list
+            vscode.window.showInformationMessage(`Target list refreshed. ${targets.length} ${targets.length === 1 ? "target": "targets"} found.`);
+            setTimeout(() => vscode.commands.executeCommand("notifications.hideToasts"), 10)
         } else {
             // Just expand the element to show its properties
             targets = element.getChildren();
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -64,7 +64,7 @@ export function activate(context: vscode.ExtensionContext) {
         }));
     context.subscriptions.push(vscode.commands.registerCommand(
         `${SETTINGS_VIEW_NAME}.refresh`,
-        async () => cdpTargetsProvider.refresh()));
+        () => cdpTargetsProvider.refresh()));
     context.subscriptions.push(vscode.commands.registerCommand(
         `${SETTINGS_VIEW_NAME}.attach`,
         (target?: CDPTarget) => {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -27,7 +27,6 @@ import {
 
 let telemetryReporter: Readonly<TelemetryReporter>;
 let browserInstance: Browser;
-let cdpTargetsProvider: CDPTargetsProvider;
 
 export function activate(context: vscode.ExtensionContext) {
     if (!telemetryReporter) {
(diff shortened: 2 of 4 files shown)


src/extension.ts
```diff
@@ -64,7 +65,11 @@ export function activate(context: vscode.ExtensionContext) {
         }));
     context.subscriptions.push(vscode.commands.registerCommand(
         `${SETTINGS_VIEW_NAME}.refresh`,
-        () => cdpTargetsProvider.refresh()));
+        async () => {
+            cdpTargetsProvider.refresh();
+            await cdpTargetsProvider.getChildren();
+            vscode.window.showInformationMessage("Refresh complete");
```
mliao95, line 71: Have you thought about adding the notification to the end of `getChildren` so that it triggers on all refreshes?  Currently this only triggers if you press the refresh button and does nothing when destroying/creating new targets.

mliao95, line 71: It would be nice if it would say the number of targets found as well.  Maybe a message such as "Refresh complete.  {num} targets found"

flynnolivia: I wasn't sure if we want to announce it when creating/destroying targets because it will produce a lot more notifications and might get cluttered (I haven't seen a way to hide notifications after an amount of time yet, lmk if you know about that). Plus, if targets are created/destroyed by opening/closing tabs in the browser, the screen reader will already announce that the tab is being open/closed.

Putting it at the end of `getChildren` would be good though because it's eventually called anyway after `cdpTargetsProvider.refresh()` (the call to `getChildren` here is redundant but helps with getting a more accurate timing for the announcement). Also, I think there might be a similar accessibility issue with the close target button, that this would solve.

flynnolivia: Good idea, I'll add that!

src/cdpTargetsProvider.ts
```diff
@@ -78,6 +78,9 @@ export default class CDPTargetsProvider implements vscode.TreeDataProvider<CDPTa
                     return a.targetJson.type < b.targetJson.type ? -1 : 1;
                 }
             });
+            // Raise and quickly hide a notification so the screen reader announces the new state of the targets list
+            vscode.window.showInformationMessage(`Target list refreshed. ${targets.length} ${targets.length === 1 ? "target": "targets"} found.`);
+            setTimeout(() => vscode.commands.executeCommand("notifications.hideToasts"), 10)
```
mliao95, line 83: I think we should remove the `notifications.hideToasts` call unless we can specify our refresh toast.  We don't want to unexpectedly remove any other toast notifications for the user.

mliao95 (approved):
LGTM after the one change.  Thanks!

src/cdpTargetsProvider.ts
flynnolivia: Ok, I'll do that.
But I looked into this a little more and found a way to make only our toast close after a timeout using `vscode.window.withProgress`. I think makes a better UX for sighted users, the only thing is that NVDA doesn't announce the message, but Narrator does. With `showInformationMessage`, both NVDA and Narrator announce the message, but the toasts can stack up and the notification list can quickly get flooded with refresh messages. Which one should we go with?

vidorteg: I took a look at the discussion at:
https://github.com/microsoft/vscode/issues/114718

and IMO I think is better to wait for those fixes to land. (We could +1 the issue by mentioning it on our bug to improve visibility). I see value on this solution but I'm not sure it will fix the issue but just temporarily mitigate it, thoughts?

vidorteg (changes requested):
left a comment

src/cdpTargetsProvider.ts
flynnolivia: I agree, I'll keep an eye on that discussion and follow up with this once the fixes land in vscode. Hopefully before July when all the accessibility bugs need to be fixed!

PR #335, opened 2021-04-01
Migrate ESLint rules from browser DevTools

flynnolivia:
I pasted the rules and overrides from the .eslintrc.js file in dt-fe and made a few small changes:
- Removed rule that interfaces cannot use the I<interface-name> naming convention
- Removed rules related to mocha
- Removed DevTools specific rules (not sure how many of these apply to the extension as well?)

After migrating and making these changes, there are 3956 errors, but 3814 are auto-fixable


mliao95 (approved):
Overall LGTM, but would want Vidal to take a look as he was looking into the ESLint rules before.  Quick question though - why do we want to deviate from browser DevTools in regards to the single/double quote?

flynnolivia:
@mliao I just noticed that it was double quotes in most of the files I've been working so far, but I don't think it's consistent throughout the codebase. This is probably something that can be auto-fixed, so it would be easy to go either way. Same for the LF/CRLF rule actually, which I learned after opening this PR!

vidorteg (approved):
Yay linting! thanks @flynnolivia

PR #346, opened 2021-04-09
Manual ESLint fixes

flynnolivia:
This PR depends on #339 which contains all of the auto-fixes. It's a lot of changes, but I checked that all the tests pass and there are no build errors. This resolves all of the current linting errors and warnings, with a few exceptions where I disabled some rules for a single line:
- Replacing occurrences of tslint-disabled rules
- Errors/warnings relating to `InspectorFrontendHostInstance.sendToVscodeOutput()` since it's usage is unconventional
- To enable ts-ignore where it's used
- A few instances that might require larger refactoring (in cdpTargetsProvider.ts, toolsHost.ts, and inspectorContentPolicy.ts)

Hopefully we can re-enable linting soon!


vidorteg (approved):
This is really amazing @flynnolivia, should we include the reenable as part of this PR?

PR #350, opened 2021-04-14
Update default start page to included startpage/index.html

bgoddar:
This PR updates the default page used when launching Microsoft Edge from about:blank to the startpage/index.html included in the repo

Closes #312

PR #351, opened 2021-04-15
Interactive documentation for side panel

### LINKED ISSUE #337: Adding interactive documentation to the extension panel
Instead of a blank document with only the instances running, we should have some docs and call to action buttons to make it easier for developers to get started. See the attached mockup for possible features.

image

flynnolivia:
This PR adds explanatory welcome content to the side pane when there are no targets to display. Users are prompted to launch an instance of Edge and configure their launch.json file to streamline the workflow.

Here's how the "getting started" workflow looks with a project that originally has no launch.json:
sidepane-cta
Summary of the changes:
- Added welcome messages and buttons
- Added 2 new commands to edit and make use of launch.json
- Updated existing tests and added tests for new functions

Closes #337


package.json
```diff
@@ -516,7 +528,39 @@
...
+                "when": "titleCommandsRegistered && launchJsonStatus == Unsupported"
+            },
+            {
+                "view": "vscode-edge-devtools-view.targets",
+                "contents": "Launch an instance of Microsoft Edge to begin inspecting and modifying your site.\n[Launch Project](command:vscode-edge-devtools-view.launchProject)",
+                "when": "titleCommandsRegistered && launchJsonStatus == Supported"
+            },
+            {
+                "view": "vscode-edge-devtools-view.targets",
+                "contents": "If you have any questions or want to learn more, check the [docs on GitHub](https://github.com/microsoft/vscode-edge-devtools/blob/master/README.md)",
```
flynnolivia, line 560: Will add a period at the end of this sentence when I go to address feedback

src/utils.ts
```diff
@@ -284,6 +290,61 @@ export async function getBrowserPath(config: Partial<IUserConfig> = {}): Promise
...
+    
+    // Create ./.vscode/launch.json if it doesn't already exist
+    const workspaceUri = vscode.workspace.workspaceFolders[0].uri;
+    fse.ensureFileSync(`${workspaceUri.fsPath}/.vscode/launch.json`);
+
+    // Append a supported debug config to their list of configurations
+    const launchJson = vscode.workspace.getConfiguration('launch', workspaceUri);
+    let configs = launchJson.get('configurations') as vscode.DebugConfiguration[];
+    let configWithInstruction = {...providedDebugConfig};
+    configWithInstruction.url += ' **Replace with your website url before launching**';
```
bgoddar, line 340: It might be a cleaner experience if we instead default to the new startpage in the generated launch.json, and leave a comment after the url that contains this text. That way, the generated launch.json is valid without any edits?

bgoddar: That would depend on #350 . Or we could default to the localhost link for now and I could update my PR if this lands first

bgoddar (changes requested):
Overall LGTM with a small request around the generated launch.json file (if it's even possible). It would also be good to get one of the others to review since I'm pretty new to this space

package.json
```diff
@@ -516,7 +528,39 @@
                     "name": "Targets"
                 }
             ]
-        }
+        },
+        "viewsWelcome": [
+            {
+                "view": "vscode-edge-devtools-view.targets",
+                "contents": "Launch an instance of Microsoft Edge to begin inspecting and modifying webpages.\n[Launch Instance](command:vscode-edge-devtools-view.launch)",
+                "when": "titleCommandsRegistered && launchJsonStatus != Supported"
```
mliao95, line 536: We should show all the content for the empty target list immediately and not wait on the `titleCommandsRegistered ` flag.  We can disable the buttons that actually launch commands until they're ready with the `enablement` flag when the commands are registered.

mliao95 (commented):
One comment about showing empty target list content right away.

flynnolivia:
I think things got a little messed up from checking in the linting changes in the middle of working on this. I'm going to open a new branch and put my changes on there.

mliao95:
> I think things got a little messed up from checking in the linting changes in the middle of working on this. I'm going to open a new branch and put my changes on there: #357

Let's close/abandon this PR if we no longer need it to avoid confusion.

PR #357, opened 2021-04-16
Interactive documentation for side panel (2)

### LINKED ISSUE #337: Adding interactive documentation to the extension panel

flynnolivia:
This PR adds explanatory welcome content to the side pane when there are no targets to display. Users are prompted to launch an instance of Edge and configure their launch.json file to streamline the workflow.

Here's how the "getting started" workflow looks with a project that originally has no launch.json:
sidepane-cta

Summary of the changes:
- Added welcome messages and buttons
- Added 2 new commands to edit and make use of launch.json
- Updated existing tests and added tests for new functions

A second attempt of #351
Closes #337

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,6 +108,18 @@
                 "command": "vscode-edge-devtools-view.viewChangelog",
                 "category": "Microsoft Edge Tools",
                 "title": "View Changelog"
+            },
+            {
+                "command": "vscode-edge-devtools-view.configureLaunchJson",
+                "category": "Microsoft Edge Tools",
+                "enablement": "titleCommandsRegistered",
+                "title": "Configure launch.json file"
+            },
+            {
+                "command": "vscode-edge-devtools-view.launchProject",
+                "category": "Microsoft Edge Tools",
+                "enablement": "titleCommandsRegistered",
+                "title": "Launch project"
             }
         ],
         "configuration": {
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -516,7 +528,38 @@
                     "name": "Targets"
                 }
             ]
-        }
+        },
+        "viewsWelcome": [
```
bgoddar, line 532: Hmm, I probably don't know enough about how extension views are registered. But, this seems like quite a bit of UI definition for this view inside the package.json. I don't see the equivalent for the list of targets, and I wonder if that's in it's own file? or handled in extensions.ts with the CDPTargetsProvider? Anyway. I can see this quickly growing in length if we add more to the sidebar view, and I wonder if there is a way to move the UI definitions to it's own file?

bgoddar: Feel free to ignore this if it doesn't make sense to do that

src/utils.ts
```diff
@@ -285,6 +291,68 @@ export async function getBrowserPath(config: Partial<IUserConfig> = {}): Promise
...
+        for (const config of configs) {
+            if (config.type === 'vscode-edge-devtools.debug' || config.type === 'msedge' || config.type === 'edge') {
+                void vscode.commands.executeCommand('setContext', 'launchJsonStatus', 'Supported');
+                return config;
+            }
+        }
+        void vscode.commands.executeCommand('setContext', 'launchJsonStatus', 'Unsupported');
+        return null;
+    }
+        void vscode.commands.executeCommand('setContext', 'launchJsonStatus', 'None');
```
bgoddar, line 320: nit, this shouldn't be indented

src/utils.ts
```diff
@@ -285,6 +291,68 @@ export async function getBrowserPath(config: Partial<IUserConfig> = {}): Promise
...
+        return null;
+
+}
+
+/**
+ * Add a template for a supported debug configuration to launch.json
+ * @returns {void}
+ */
+export async function configureLaunchJson(): Promise<void> {
+    if (!vscode.workspace.workspaceFolders)
```
bgoddar, line 330: I wonder if we should find a way to handle this case more gracefully for users. Would they ever be in a scenario where they try to create a launchJson without opening a workspace folder? would they get an error message if this fails or would it just be silent? I could see the silent fail case being confusing for users if they hit it.

flynnolivia: I don't think that scenario should be possible because in package.json, the button to create a launch.json is only shown if there is a workspace folder open. The if statement was needed to quiet a typescript error about `workspaceFolders` possibly being undefined. We'll need to be careful of that if we use this function in more places later on, so I guess we might as well put in an error message now.

package.json
flynnolivia: I think we will need to keep it in package.json because viewsWelcome is a contribution point specifically defined for VSCode extensions (more about that [here](https://code.visualstudio.com/api/references/contribution-points#contributes.viewsWelcome))

Quick rundown though: the sidepane view is registered in package.json as `vscode-edge-devtools-view.targets`, then in extension.js the sidepane view subscribes to CDPTargetsProvider, which has the logic to get whatever info VSCode needs to display the targets. viewWelcome only shows when the list that it's view subscribes to is empty, plus some additional conditions (in the `"when"` property). So there's a lot more UI definition for viewsWelcome vs the targets view in package.json because viewsWelcome is more static and the target list is more dynamic.

bgoddar: Ahh, totally makes sense. Thanks for the explanation and docs!

mliao95:
One edge case NIT:

Could we refresh the sidepanel view upon refresh?  Here is a case with some odd behavior:
launch-config-refresh

mliao95 (approved):
Overall LGTM after the comment above.

PR #365, opened 2021-04-20
prototype debug share

bgoddar:
Very rough WIP that connects devtools to the active VSCode jsdebug target using CDP share

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -47,6 +47,10 @@ export function activate(context: vscode.ExtensionContext): void {
         void launch(context);
     }));
 
+    context.subscriptions.push(vscode.commands.registerCommand(`${SETTINGS_STORE_NAME}.attachShare`, (): void => {
+        void attach(context);
+    }));
+
     // Register the launch provider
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -139,6 +149,59 @@ export function activate(context: vscode.ExtensionContext): void {
...
+        }
+
+        let targetWebsocketUrl;
+        try {
+            const addr: any = await vscode.commands.executeCommand(
+            'extension.js-debug.requestCDPProxy',
+            session.id,
+            );
+            const formed = `ws://${addr.host}:${addr.port}`;
+            vscode.window.showInformationMessage(`Address copied to your clipboard (${formed})`);
```
mliao95, line 180: Does the address get written in the keyboard as a part of `extension.js-debug.requestCDPProxy`?  It isn't a big deal, but not a fan of buttons unexpectedly overwriting a user's clipboard.

bgoddar:
replaced by https://github.com/microsoft/vscode-edge-devtools/pull/391

PR #366, opened 2021-04-22
Streamlined workflow for launching localhost

flynnolivia:
This PR adds two workflows for automatically adding configs to launch.json that causes a localhost watch server to be launched in parallel with launching Edge and attaching a project. It also updates tests to account for new functionality.

(click to view an enlarged version)
Workflow 1: Generating a new launch.json with the appropriate configs (launch watch server, launch+attach Edge, and a compound config to execute them in parallel)
watch-server-new-launchjson

Workflow 2: Adding configs to an existing launch.json that already has a supported debug config (launch watch server and compound config)
watch-server-existing-launchjson


mliao95:
I'm getting this error when trying to start the localhost server:
image

I tried running `npm install`, but didn't work.  Have you seen this error before?  If it isn't specific to my local setup the extension dependency might not be set up completely.

@bgoddar - could you check to see if this runs for you?

bgoddar:
> I'm getting this error when trying to start the localhost server:
> image
> 
> I tried running `npm install`, but didn't work. Have you seen this error before? If it isn't specific to my local setup the extension dependency might not be set up completely.
> 
> @bgoddar - could you check to see if this runs for you?

This seems to work fine for me, but I' was using it inside a project that already had the liveserver dependency. Let me try a different folder

bgoddar:
> > I'm getting this error when trying to start the localhost server:
> > image
> > I tried running `npm install`, but didn't work. Have you seen this error before? If it isn't specific to my local setup the extension dependency might not be set up completely.
> > @bgoddar - could you check to see if this runs for you?
> 
> This seems to work fine for me, but I' was using it inside a project that already had the liveserver dependency. Let me try a different folder

Yeah, best I can tell, this works if live-server is already on you package.json? @flynnolivia can you confirm if this works on a fresh project with just some html/css? or if it needs live-server already configured in the project?

flynnolivia:
> > > I'm getting this error when trying to start the localhost server:
> > > image
> > > I tried running `npm install`, but didn't work. Have you seen this error before? If it isn't specific to my local setup the extension dependency might not be set up completely.
> > > @bgoddar - could you check to see if this runs for you?
> > 
> > 
> > This seems to work fine for me, but I' was using it inside a project that already had the liveserver dependency. Let me try a different folder
> 
> Yeah, best I can tell, this works if live-server is already on you package.json? @flynnolivia can you confirm if this works on a fresh project with just some html/css? or if it needs live-server already configured in the project?

Yeah it works on my end with a simple project. I got the same error when I was first tried to switch from the reload package -> live-server but it worked after I installed live-server globally for myself. So I thought it should work if we have it as a dependency for the extension 🤔 Any idea what needs to be done in order to finish setting up the dependency?

mliao95:
> > > > I'm getting this error when trying to start the localhost server:
> > > > image
> > > > I tried running `npm install`, but didn't work. Have you seen this error before? If it isn't specific to my local setup the extension dependency might not be set up completely.
> > > > @bgoddar - could you check to see if this runs for you?
> > > 
> > > 
> > > This seems to work fine for me, but I' was using it inside a project that already had the liveserver dependency. Let me try a different folder
> > 
> > 
> > Yeah, best I can tell, this works if live-server is already on you package.json? @flynnolivia can you confirm if this works on a fresh project with just some html/css? or if it needs live-server already configured in the project?
> 
> I had the same error when I first tried to switch from the reload package -> live-server but it worked after I installed live-server globally for myself. So I thought it should work if we have it as a dependency for the extension 🤔
> 
> I just uninstalled it from my global package.json and got the same error, but when live-server is actually installed, it should work with a simple html/css project. Any idea what needs to be done in order to finish setting up the dependency?

The `launch.json` command is just telling the user's terminal to run `live-server`.  Although the Live Server extension is a dependency of our extension and is accessible within our extension, that doesn't mean we download it globally on the user's machine.

I'm thinking we have to create commands to launch and manage the live server from within our extension and call that command within the `launch.json`.  Might be worth investigating to see how much work that would be. Given that we would need to add UI to manage the server, we should consider if this is a necessary change before committing the time to finishing.

flynnolivia:
> > > > > I'm getting this error when trying to start the localhost server:
> > > > > image
> > > > > I tried running `npm install`, but didn't work. Have you seen this error before? If it isn't specific to my local setup the extension dependency might not be set up completely.
> > > > > @bgoddar - could you check to see if this runs for you?
> > > > 
> > > > 
> > > > This seems to work fine for me, but I' was using it inside a project that already had the liveserver dependency. Let me try a different folder
> > > 
> > > 
> > > Yeah, best I can tell, this works if live-server is already on you package.json? @flynnolivia can you confirm if this works on a fresh project with just some html/css? or if it needs live-server already configured in the project?
> > 
> > 
> > I had the same error when I first tried to switch from the reload package -> live-server but it worked after I installed live-server globally for myself. So I thought it should work if we have it as a dependency for the extension 🤔
> > I just uninstalled it from my global package.json and got the same error, but when live-server is actually installed, it should work with a simple html/css project. Any idea what needs to be done in order to finish setting up the dependency?
> 
> The `launch.json` command is just telling the user's terminal to run `live-server`. Although the Live Server extension is a dependency of our extension and is accessible within our extension, that doesn't mean we download it globally on the user's machine.
> 
> I'm thinking we have to create commands to launch and manage the live server from within our extension and call that command within the `launch.json`. Might be worth investigating to see how much work that would be. Given that we would need to add UI to manage the server, we should consider if this is a necessary change before committing the time to finishing.

Ok, that makes sense. I'll do some more investigation into managing the live server from within our extension, but otherwise hold off on this until we get a more concrete idea of how we want the server launching experience to be.

PR #367, opened 2021-04-22
Remove iframe and host DevTools directly inside webview

### LINKED ISSUE #359: Updates to resource loading in VSCode Insiders breaks the extension
VSCode is switching from custom resource loading using `vscode-webview-resource:` to using service workers to load resources. This means that resources will now be loaded into a structure like `<iframe><iframe></iframe></iframe>` instead of `<webview><iframe></iframe></webview>`.

To repro: Within VSCode Insiders, use the extension to launch an instance of edge
Expected result: The devtools appear in a new pane to the right
Actual result: The pane that typically holds the devtools is blank, except for an error message like `ResourceNotFoundThe specified resource does not exist. RequestId:68872f37-801e-0011-5c5c-353d11000000 Time:2021-04-19T20:43:37.3899983Z`

Commits containing the breaking changes:
https://github.com/microsoft/vscode/commit/de9887d9e0eaf402250d2735b3db5dc340184b74
https://github.com/microsoft/vscode/commit/d05ded6d3b64fed4a3cc74106f9b6c72243b18de

mliao95:
Issue:
- VS Code deprecated loading local resources inside iframe in VS Code version 1.56.0-insider

Changes:
- Load all resources from `inspector.html` directly inside the webview

Steps:
- Migrate `inspector.html` components into webview html in `devtoolsPanel.ts`
- Fix messaging after removing iframe
    - remove `messaging.ts` and `mainMessaging.ts` since we no longer need to handle iframe/webview communication
    - use `acquireVsCodeApi` and `vscode.postMessage` to communicate from the webview to vscode
        - VS Code to webview = `this.panel.webview.postMessage(...)`, webview to VS Code = `vscode.postMessage(...)`
- Enable web socket connection inside DevTools
    - fixed via  `applyQueryParamsObjectPatch`
    - DevTools uses URL parameters like `?ws=trueD&experiments=true&edgeThemes=true` to set flags and enable websocket debugging.
        - This was originally passed in with `inspector.html` but since we no longer reference that file, we patch it into the creation of the queryParams object.

closes #359


src/host/host.ts
```diff
@@ -58,8 +61,6 @@ export function initialize(dtWindow: IDevToolsWindow): void {
         // Override the resource loading once the window has loaded so that we can control it
         const resourceLoader = ToolsResourceLoader.overrideResourceLoading(dtWindow.Root.Runtime);
         dtWindow.InspectorFrontendHost.setResourceLoader(resourceLoader);
-
-        dtWindow.importScriptPathPrefix = dtWindow.importScriptPathPrefix.replace('null', 'vscode-webview-resource:');
```
mliao95, line 62: @vidorteg  - we no longer need this, right?

src/devtoolsPanel.ts
```diff
@@ -298,31 +298,44 @@ export class DevToolsPanel {
     }
 
     private getHtmlForWebview() {
-        const htmlPath = vscode.Uri.file(path.join(this.extensionPath, 'out/tools/front_end', 'inspector.html'));
-        const htmlUri = this.panel.webview.asWebviewUri(htmlPath);
+        // stringsUri and inspectorUri are the files that used to be loaded in inspector.html, we are trying to load them directly to the webview here.
```
bgoddar, line 301: nit: wdyt about adding links to the relevant vscode commits that are forcing this change?

src/devtoolsPanel.ts
```diff
@@ -298,31 +298,44 @@ export class DevToolsPanel {
...
+        const stringsPath = vscode.Uri.file(path.join(this.extensionPath, 'out/tools/front_end', 'strings.js'));
+        const stringsUri = this.panel.webview.asWebviewUri(stringsPath);
+
+        const inspectorPath = vscode.Uri.file(path.join(this.extensionPath, 'out/tools/front_end', 'inspector.js'));
+        const inspectorUri = this.panel.webview.asWebviewUri(inspectorPath);
 
         const scriptPath = vscode.Uri.file(path.join(this.extensionPath, 'out', 'host', 'messaging.bundle.js'));
```
bgoddar, line 308: looks like the messaging files were deleted? should this be deleted too?

src/devtoolsPanel.ts
```diff
@@ -298,31 +298,44 @@ export class DevToolsPanel {
...
+                    script-src 'self' 'unsafe-eval' ${this.panel.webview.cspSource};
                     frame-src ${this.panel.webview.cspSource};
-                    script-src ${this.panel.webview.cspSource};
-                    style-src ${this.panel.webview.cspSource};
-                    ">
+                    connect-src 'self' ${this.panel.webview.cspSource};
+                ">
+                <meta name="referrer" content="no-referrer">
                 <link href="${stylesUri}" rel="stylesheet"/>
                 <script src="${scriptUri}"></script>
```
bgoddar, line 333: same comment as above about deleting the messaging bundle

src/devtoolsPanel.ts
bgoddar: Probably explains the double vscode api loading we saw

src/host/host.ts
bgoddar: IIUC, this should be safe to remove, as it was only needed to transform loading requests from inside the iframe into a vscode webview resource, but that should all be done automatically now since we are in the same webview/domain as the resource loader

bgoddar (commented):
I've done a very shallow test of functionality on stable and insiders and this appears to be working so far. Couple of comments

bgoddar:
I'm also seeing svg icons fail to load 😢 since it makes the dom tree really hard to use without them, we should probably figure out what's going on here
image

src/devtoolsPanel.ts
mliao95: I think that is reasonable.  Added.

flynnolivia:
Generally lgtm, seems to be working normally again on my end as far as I can tell!
Just want to make sure I understand correctly: The main issue was putting the devtools inside an iframe, and before we put the devtools in an iframe inside a webview. So now the devtools are only inside a webview, and we don't need those messaging classes anymore because there's one less layer of messaging to handle?

vidorteg (changes requested):
Hey @mliao95 

I think this change is big enough to require a design discussion meeting. Could you set one please? Some of the things I would like to understand better is how our current design changed from :
image

mliao95:
> Generally lgtm, seems to be working normally again on my end as far as I can tell!
> Just want to make sure I understand correctly: The main issue was putting the devtools inside an iframe, and before we put the devtools in an iframe inside a webview. So now the devtools are only inside a webview, and we don't need those messaging classes anymore because there's one less layer of messaging to handle?

Pretty much.  To dig a little deeper, the core issue was loading local resources inside an iframe inside a webview.  For instance, loading resources from a website or localhost would actually still work. In general, VS Code cautions against using iframes in extensions since it causes problems.

The original messaging lived inside the webview outside the iframe and would pass messages between VS Code and the iframe. The iframe would pass messages into it via `window.parent.postMessage` and VS Code would pass messages via `this.panel.webview.postMessage`.  Now that we removed the iframe, we just tell the DevTools to post messages directly to VS Code using the `vscode.postMessage` call and the DevTools can get the message directly when VS Code calls `this.panel.webview.postMessage`.

bgoddar (approved):

vidorteg (approved):

PR #370, opened 2021-04-28
Telemetry for empty target list view

flynnolivia:
We shouldn't check this in until we get access to the Kusto cluster and verify that it all works, but we can use this PR to further discuss how we want the event structure to be.
@mliao95 I tried to implement the structure you were telling me about on Monday.

This PR adds:
- An enum that maps UI button descriptions to number codes
- Telemetry events for launching a browser instance, launching a project using launch.json, viewing docs on github, and generating/configuring launch.json
- A command to view the docs in order to add the telemetry event (before it was a link associated with the button in package.json)
- Updated tests


src/utils.ts
```diff
@@ -111,6 +111,15 @@ export interface IRemoteTargetJson {
...
 }
 
+/** enum {string} */
+export const buttonCode: Record<string, string> = {
+    launchBrowserInstance: '0',
+    launchProject: '1',
+    viewDocumentation: '2',
+    configureLaunchJson: '3',
+    generateLaunchJson: '4',
+};
```
mliao95, line 121: Nice!  This is exactly what I had in mind.

src/utils.ts
```diff
@@ -293,13 +302,13 @@ export async function getBrowserPath(config: Partial<IUserConfig> = {}): Promise
 
 /**
  * Gets a supported debug config and updates the status of the launch.json file associated with the current workspace
- * @returns {vscode.DebugConfiguration | null}
+ * @returns {vscode.DebugConfiguration | string}
  */
- export function getLaunchJson(): vscode.DebugConfiguration | null {
+ export function getLaunchJson(): vscode.DebugConfiguration | string {
```
mliao95, line 307: I don't love using different return types to indicate state, but I can see how we would need to.  Let us at least scope in what values are acceptable instead of a generic string return type.

I was thinking we could use a type and do something like this:
```
type LaunchConfig = 'None' | 'Unsupported' | vscode.DebugConfiguration;
...
export function getLaunchJson(): LaunchConfig { ...
```

src/extension.ts
```diff
@@ -124,18 +126,26 @@ export function activate(context: vscode.ExtensionContext): void {
...
             void configureLaunchJson();
             launchConfig = getLaunchJson();
         }));
     context.subscriptions.push(vscode.commands.registerCommand(
         `${SETTINGS_VIEW_NAME}.launchProject`,
         () => {
+            telemetryReporter.sendTelemetryEvent('user/buttonPress', { 'VSCode.buttonCode': buttonCode.launchProject });
             launchConfig = getLaunchJson();
-            if (vscode.workspace.workspaceFolders && launchConfig) {
+            if (vscode.workspace.workspaceFolders && typeof launchConfig === 'object') {
```
mliao95, line 140: Let's set `typeof launchConfig === 'object'` to a const with a meaningful name to improve readability.  A suggestion would be something like this:
`const isValidLaunchConfig = typeof launchConfig === 'object';`

mliao95 (approved):
Love the change!  Have some quality of code suggestions, but will sign off after those are addressed.

PR #377, opened 2021-05-05
[Telemetry]: Adding telemetry to track if the target URL is localhost or not

flynnolivia:
Changes in this PR:
- Adds a function `reportUrlType()` that determines whether the url is localhost, file:// protocol, or other, and sends this property with the telemetry event
- `reportUrlType()` is called when the browser is launched (except if it's launching the start page)
- `reportUrlType()` is called each time a target's url changes if the target is a page
- Event listeners on the browser instance are created with the `on()` function now, since `addListener()` is deprecated


src/extension.ts
```diff
@@ -289,16 +290,29 @@ export async function launch(context: vscode.ExtensionContext, launchUrl?: strin
...
         browserInstance = await launchBrowser(browserPath, port, url, userDataDir);
-        browserInstance.addListener('targetcreated', () => {
+        if (url !== SETTINGS_DEFAULT_URL) {
+            reportIsLocalhost(url);
+        }
+        browserInstance.on('targetcreated', () => {
             cdpTargetsProvider.refresh();
         });
-        browserInstance.addListener('targetdestroyed', () => {
+        browserInstance.on('targetdestroyed', () => {
```
mliao95, line 299: What are the functional differences of swapping out the `addListener` with `on`?

mliao95 (approved):
Overall LGTM with one question.

src/extension.ts
mliao95: Ah, just saw the deprecated note on `addListener`

src/extension.ts
```diff
@@ -289,16 +290,29 @@ export async function launch(context: vscode.ExtensionContext, launchUrl?: strin
...
         await attach(context, url, config);
     }
 }
 
 export function setLaunchConfig(): void {
     launchConfig = getLaunchJson();
 }
+
+function reportIsLocalhost(url: string): void {
+    const isLocalhost = url.includes('localhost:').toString();
```
vidorteg, line 316: Have we considered the case where a local address is used in a different way? e.g  127.0.0.1 or any other localhost vairant( pointing directly to local ip 192.x.x.x).  (AFAIK) We might have to inspect the request to check whether it is coming from a local server.

vidorteg (changes requested):
Left a comment

src/extension.ts
flynnolivia: This is a good catch. I'm not sure what we can check on the request besides the url, but I'd like to hear more about that if you think it's the route we should go.
I was looking at the different reserved address blocks [here](https://en.wikipedia.org/wiki/Reserved_IP_addresses), could we just check whether the url is in the localhost block or one of the blocks with the description "Used for local communications within a private network"?

vidorteg: spend sometime around this and I'm not sure if there is an easy way that covers all bases. So we can probably get around this by verifying at least a couple of scenarios:
- localhost (already covered, just make it case insensitive)
- private subnets (😶does not covers all scenarios but most)
- protocol (is file:// considered localhost)?

Thoughts on this?

flynnolivia: I think this should cover the most common cases! I'm not sure about the file:// part though.
Taking a step back, my understanding of this telemetry point is that it should tell us if many people are using localhost or 127.0.0.1, so that we can decide whether to invest time into providing a way to start a localhost server from our extension. In that case, I'm not sure we need to worry about the private subnets, file://, or other types of local addresses, since AFAIK the only server we'll be spinning up from within our extension is localhost/127.0.0.1.

@mliao95 any thoughts, and can you verify the question this telemetry point is intended to answer?

mliao95: I believe @codepo8 suggested this bit of telemetry, so he can speak more to the reasoning behind this.

We all agree that localhost and 127.0.0.1 should be accounted for, so the real discussion is really about private subnets and file://. In my opinion, I don't believe the private subnets are needed since I would assume they have their own separate workflow, and if they wanted or needed to launch via localhost, they would have.  

The file:// is a little more interesting.  I can see a developer that doesn't know how to start a local server debug a file by just opening it in the browser.  In that case, a high number of file:// users could also indicate a need of an easy to understand way to spin up a local server.  However, I'm not sure if we should combine the file:// telemetry with the localhost telemetry.  I think we should differentiate it using properties or a different event.

flynnolivia: I think it makes sense to include the file:// telemetry in this event, but differentiate whether the url is file:// vs localhost vs other using properties. It could be a useful success metric to see if the amount of people using file:// changes after we introduce the localhost workflow. But, I'll wait and see what Chris has to say before moving forward with that.

vidorteg: > I think it makes sense to include the file:// telemetry in this event, but differentiate whether the url is file:// vs localhost vs other using properties. 

Good idea, +1.

For the subnet I think we can consider any url in the form [0-255].[0-255].[0-255].[0-255] a local address, in my experience it is extremely rare to use the ip address directly to connect to server in a production environment. Reason I think we should cover private subnets is because (AFAIK) Linux does not have localhost by default (you only have 127.0.0.1( so it is a common practice (in my experience) to access a local server using the ip address + port. (Not sure if this is true for Mac)

vidorteg (approved):
LGTM

PR #381, opened 2021-05-06
Fixing a race condition on cdpTargetsProvider.ts

vidorteg:
Issue:
image
I hit this a couple of time. There is a race condition when the favicon icon is cleared. 

Fix:
- Simplified the code
- There were a couple of issues when we were triggering the deletes & downloads, instead of triggering it on every target that gets launched it triggers only when the target list changes. (most of the refactor is due to this)
- Another issues was  mixing of events and promises when reading directory and when deleting the files. I fixed it using fs.promises instead and took out the events.


flynnolivia (approved):
LGTM! Maybe want to wait for Michael's review to be sure though, looks like he did this feature

mliao95 (changes requested):
Overall I think this is the right change.  I just ran into a bug that I think we should fix in this PR.
1. Launch instance
2. Navigate to amazon.com
3. Open vscode-edge-devtools\resources\favicons in explorer
4. Close out of browser
5. Refresh target list

Expected Result:
all .ico files are deleted

Actual Result:
none of the .ico files are deleted

This is because of the `amazon-adsystemFavicon.ico` and `ssl-images-amazonFavicon.ico` which is an attempt to download the favicon from amazon iframes. Generally, downloading favicons from iframes are fine, but for some reason these are stuck downloading.

Here are some changes that would help this issue:
- If we run into an instance where we cannot delete a file, we should move onto the next file instead of halting there.
- We should have a timeout for `downloadFaviconFromSitePromise` so that we don't run into this issue.

mliao95 (approved):
LGTM.

PR #390, opened 2021-05-11
[Accessibility]: Add success alert when selecting 'Toggle property and continue editing selected'

mliao95:
The accessibility issue is that screen reader users do not hear the success message of the "Toggle property and continue editing selected" context menu option for styles properties:

image

This is in part due to the same context menu/webview issue that was causing the accessibility bug here: https://github.com/microsoft/vscode-edge-devtools/pull/332

This PR adds an alert call inside the patch created in the PR above so that it will announce the success message.


flynnolivia (approved):
Tested it out, lgtm!

vidorteg (approved):
LGTM

PR #391, opened 2021-05-14
Add ability to attach to active debug target using shared CDP connection

bgoddar:
This Pull request registers a new command for the extension that allows VSCode's JavaScript Debugger to attach the devtools to the currently active debug target.

It also leverages the new CDPProxy that allows our extension to connect to the client using the same websocket as the JsDebugger, giving us the benefits of:
- Appearing as a single debugger to the debug client page
- Allowing us to attach to debugging instances of Edge that are connected via annonymous pipes from VSCode debugger, instead of only remote-debugging-port launched Edge instances

image

image

image

Testing instructions:
To test the new features, Follow the instructions to install the Nightly flight of the JavaScript Debugger into VSCode Insiders from: https://github.com/microsoft/vscode-js-debug#nightly-extension
Then, debug a build from this branch.
In the resulting Extension host environment:
1. Attach the JsDebugger to a target page using an Edge launch.json:
        {
            "name": "Launch Edge",
            "request": "launch",
            "type": "pwa-msedge",
            "url": "http://localhost:8080",
            "webRoot": "${workspaceFolder}"
        },

2. Click the new "inspect icon" in the Vscode debugger bar
2a. An alternative is to use the 'Debug: Open Browser Devtools' command prompt

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,16 @@
                 }
             },
             {
+                "command": "vscode-edge-devtools-view.attachShare",
+                "category": "Microsoft Edge Tools",
+                "title": "Attach to active debug target",
+                "enablement": "titleCommandsRegistered",
+                "icon": {
+                    "light": "resources/light/launch.svg",
+                    "dark": "resources/dark/launch.svg"
+                }
+            },
+            {
                 "command": "vscode-edge-devtools-view.refresh",
                 "category": "Microsoft Edge Tools",
                 "title": "Refresh Targets",
@@ -457,6 +467,10 @@
                     "when": "view == vscode-edge-devtools-view.targets"
                 },
                 {
+                    "command": "vscode-edge-devtools-view.attachShare",
+                    "when": "view == vscode-edge-devtools-view.targets"
+                },
+                {
                     "command": "vscode-edge-devtools-view.refresh",
                     "when": "view == vscode-edge-devtools-view.targets"
                 },
@@ -476,6 +490,11 @@
             "view/title": [
                 {
                     "command": "vscode-edge-devtools-view.launch",
+                    "when": "view == vscode-edge-devtools-view.targets",
+                    "group": "navigation"
+                },
+                {
+                    "command": "vscode-edge-devtools-view.attachShare",
                     "when": "view == vscode-edge-devtools-view.targets",
                     "group": "navigation"
                 },
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "activationEvents": [
         "onCommand:vscode-edge-devtools.attach",
         "onCommand:vscode-edge-devtools.launch",
+        "onCommand:vscode-edge-devtools.attachToCurrentDebugTarget",
         "onWebviewPanel:vscode-edge-devtools",
         "onView:vscode-edge-devtools-view.targets",
         "onDebug"
@@ -55,19 +56,14 @@
                 "category": "Microsoft Edge Tools"
             },
             {
+                "command": "vscode-edge-devtools.attachToCurrentDebugTarget",
+                "category": "Microsoft Edge Tools",
+                "title": "Attach to active debug target"
+            },
+            {
                 "command": "vscode-edge-devtools-view.launch",
                 "category": "Microsoft Edge Tools",
                 "title": "Open a new tab",
-                "enablement": "titleCommandsRegistered",
-                "icon": {
-                    "light": "resources/light/launch.svg",
-                    "dark": "resources/dark/launch.svg"
-                }
-            },
-            {
-                "command": "vscode-edge-devtools-view.attachToCurrentDebugTarget",
-                "category": "Microsoft Edge Tools",
-                "title": "Attach to active debug target",
                 "enablement": "titleCommandsRegistered",
                 "icon": {
                     "light": "resources/light/launch.svg",
@@ -467,10 +463,6 @@
                     "when": "view == vscode-edge-devtools-view.targets"
                 },
                 {
-                    "command": "vscode-edge-devtools-view.attachToCurrentDebugTarget",
-                    "when": "view == vscode-edge-devtools-view.targets"
-                },
-                {
                     "command": "vscode-edge-devtools-view.refresh",
                     "when": "view == vscode-edge-devtools-view.targets"
                 },
@@ -490,11 +482,6 @@
             "view/title": [
                 {
                     "command": "vscode-edge-devtools-view.launch",
-                    "when": "view == vscode-edge-devtools-view.targets",
-                    "group": "navigation"
-                },
-                {
-                    "command": "vscode-edge-devtools-view.attachToCurrentDebugTarget",
                     "when": "view == vscode-edge-devtools-view.targets",
                     "group": "navigation"
                 },
--- a/package.json
+++ b/package.json
@@ -602,7 +602,9 @@
         "download-edge": "node scripts/downloadAndExtractEdge.js"
     },
     "dependencies": {
+        "bufferutil": "4.0.1",
         "puppeteer-core": "9.1.1",
+        "utf-8-validate": "5.0.2",
         "vscode-chrome-debug-core": "6.8.9",
         "vscode-extension-telemetry": "0.1.7",
         "ws": "7.4.5",
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
                 }
             },
             {
-                "command": "vscode-edge-devtools-view.attachShare",
+                "command": "vscode-edge-devtools-view.attachToCurrentDebugTarget",
                 "category": "Microsoft Edge Tools",
                 "title": "Attach to active debug target",
                 "enablement": "titleCommandsRegistered",
@@ -467,7 +467,7 @@
                     "when": "view == vscode-edge-devtools-view.targets"
                 },
                 {
-                    "command": "vscode-edge-devtools-view.attachShare",
+                    "command": "vscode-edge-devtools-view.attachToCurrentDebugTarget",
                     "when": "view == vscode-edge-devtools-view.targets"
                 },
                 {
@@ -494,7 +494,7 @@
                     "group": "navigation"
                 },
                 {
-                    "command": "vscode-edge-devtools-view.attachShare",
+                    "command": "vscode-edge-devtools-view.attachToCurrentDebugTarget",
                     "when": "view == vscode-edge-devtools-view.targets",
                     "group": "navigation"
                 },
--- a/package.json
+++ b/package.json
@@ -56,11 +56,6 @@
                 "category": "Microsoft Edge Tools"
             },
             {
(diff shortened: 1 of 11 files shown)


src/panelSocket.ts
```diff
@@ -13,11 +13,13 @@ export class PanelSocket extends EventEmitter {
     private socket: WebSocket | undefined;
     private isConnected = false;
     private messages: string[] = [];
```
ngaitonde, line 15: ; missing. Otherwise PR looks good to me. Signing off.

ngaitonde (approved):

src/panelSocket.ts
```diff
@@ -74,14 +76,17 @@ export class PanelSocket extends EventEmitter {
         this.socket.onmessage = ev => this.onMessage(ev);
         this.socket.onerror = () => this.onError();
         this.socket.onclose = () => this.onClose();
+
```
mliao95, line 79: We can probably remove this newline

src/extension.ts
```diff
@@ -159,6 +163,44 @@ export function activate(context: vscode.ExtensionContext): void {
...
+    let targetWebsocketUrl;
+    try {
+        const addr: any = await vscode.commands.executeCommand(
+        'extension.js-debug.requestCDPProxy',
+        session.id,
+        );
+        const formed = `ws://${addr.host}:${addr.port}${addr.path || ''}`;
+        targetWebsocketUrl = formed;
+    } catch (e) {
+        vscode.window.showErrorMessage(e.message);
```
mliao95, line 190: We don't necessarily need to do this in this PR, but we should start utilizing `sendTelemetryErrorEvent` to record failures that happen within the extension code.  Currently, we only use `sendTelemetryErrorEvent` for DevTools errors.

Feel free to send error telemetry in the places where you show the error toast notification, but otherwise we can change this when we make a sweep of the extension code to implement better error telemetry.

src/panelSocket.ts
```diff
@@ -113,4 +118,24 @@ export class PanelSocket extends EventEmitter {
...
+                "CSS.*",
+                "DOMDebugger.*",
+                "Network.*",
+                "Page.*",
+                "Target.*",
+                "Overlay.*"
+                ]
+            }
+        }
+        this.socket?.send(JSON.stringify(registrationMessage));
```
mliao95, line 139: NIT:  I know in the DevTools we tend to avoid optional chaining with `?`.  I forget the exact reason, but it is something along the lines of it complicating and abstracting error detection.

That being said, this usage of it seems safe, but this is more of a question of if we want to establish a code-wide standard like we do with the DevTools.

mliao95 (commented):
So far it looks good! Thanks for the change! 

Left a couple minor comments.  I'll take another, more thorough look once the build is green.

src/panelSocket.ts
bgoddar: Good point. I think the standard practice in the extension so far has also been to avoid `?.`, so I'll update this usage

bgoddar:
A bit more context on the weird case of missing events: I can fairly consistenly repro by:
1) Attaching a debug session of js-debugger
2) running the command to attach devtools from Ctrl+shift+P
3) Closing devtools
4) Reopening devtools.

The issue then resolves if I reload the page. I'm not sure if it's something weird on our end about remembering an old session? Or if there is an issue in the js-debugger side of things

src/extension.ts
```diff
@@ -249,6 +259,53 @@ export async function attach(
...
+        const addr: RequestCDPProxyResult|undefined = await vscode.commands.executeCommand(
+        'extension.js-debug.requestCDPProxy',
+        session.id,
+        );
+        if (addr) {
+            const formed = `ws://${addr.host}:${addr.port}${addr.path || ''}`;
+            targetWebsocketUrl = formed;
+        }
+    } catch (e) {
+        if (e instanceof Error) {
```
vidorteg, line 289: why is this needed? any other exception will be swallowed by the return statement.

src/extension.ts
```diff
@@ -249,6 +259,53 @@ export async function attach(
     } while (useRetry && Date.now() - startTime < timeout);
 }
 
+export async function attachToCurrentDebugTarget(context: vscode.ExtensionContext): Promise<void> {
```
vidorteg, line 262: I think is better to have the logic described here in it's own class and factory (CDPDebugTarget && CDPDebugTargetProvider), that will help with encapsulation.

src/panelSocket.ts
```diff
@@ -113,4 +117,26 @@ export class PanelSocket extends EventEmitter {
 
         this.isConnected = false;
     }
+
+    private registerForCDPEvents() {
```
vidorteg, line 121: nit: registerForJsDebugDCPSharedEvents, as the behavior is specific for JSDebug

src/panelSocket.ts
```diff
@@ -13,11 +13,13 @@ export class PanelSocket extends EventEmitter {
     private socket: WebSocket | undefined;
     private isConnected = false;
     private messages: string[] = [];
+    private isCDPShared = false;
 
-    constructor(targetUrl: string, postMessageToDevTools: IDevToolsPostMessageCallback) {
+    constructor(targetUrl: string, postMessageToDevTools: IDevToolsPostMessageCallback, isCDPShared = false) {
```
vidorteg, line 18: nit: IsJsDebugCDPShared, as the behavior is specific for JSDebug

src/panelSocket.ts
```diff
@@ -113,4 +117,26 @@ export class PanelSocket extends EventEmitter {
...
+                'Network.*',
+                'Page.*',
+                'Target.*',
+                'Overlay.*',
+                ],
+            },
+        };
+        if (this.socket) {
+            this.socket.send(JSON.stringify(registrationMessage));
+        }
```
vidorteg, line 140: I'm a little confused by what we are doing in here. AFAIK this module represents the "real-socket" we use to establish the connection with the browser, so not sure what is the role of JsDebug.subscribe, would you mind adding some details on the expected workflow in the description?

vidorteg (changes requested):
left some comments.

src/extension.ts
bgoddar: Typescipt gets mad if you access e.message without first checking if it's an Error instead of Any

src/panelSocket.ts
bgoddar: In the case of sharing CDP connection, this socket is provided by jsDebug, not the browser, and everything is proxied. By default, using the shared JsDebug proxy does not send any events unless specifically subscribed to. More info is available here: https://github.com/connor4312/cdp-proxy-requestor. I'll add docs here as well before checking in

src/extension.ts
bgoddar: I feel like that might be overkill? the cdpTarget and cdpTargetsprovider need to handle finding and providing several targets available from the browser of different type, and handles the sidebar ui, but this logic is the same every time, and is shorter than the launch/attach workflow, and has no UI components. I would basically just be moving this function into a separate file, but would still need to call/register it here

src/panelSocket.ts
vidorteg: thanks for details, I think I get it what it is happening now. The approach seems the correct one to me but I think we need two classes for this:
- PanelSocket: the usual socket we know.
- DebugProxyPanelSocket: the specific implementation, (could extend PanelSocket for code reusing code)

That will help if the behavior fork in the future

bgoddar: Good call. I've created a new debug proxy socket class. Helped get rid of a bunch of passing booleans into contructors

src/extension.ts
vidorteg: got it, that makes sense. My only suggestion would be to rethrow to raise unexpected exceptions.
`if (e instanceof Error) {
            telemetryReporter.sendTelemetryErrorEvent('command/attachToCurrentDebugTarget/devtools', {message: e.message});
            void vscode.window.showErrorMessage(e.message);
        } else{
            throw e;
        }`

src/extension.ts
vidorteg: my bad, forgot to delete this one after the clarification on the other comment 👍 (I thought we were handling the connection through different sockets)

src/JsDebugProxyPanelSocket.ts
```diff
@@ -0,0 +1,42 @@
...
+    protected onOpen(): void {
+        if (this.socket) {
+            this.registerForJsDebugSharedCDPEvents();
+        }
+        super.onOpen();
+    }
+
+    private registerForJsDebugSharedCDPEvents() {
+        // register for custom events from jsdebug:
+        // TODO before checkin: add docs here
```
mliao95, line 22: Don't forget to either finish or update this TODO

mliao95 (approved):
LGTM

src/extension.ts
```diff
@@ -249,6 +259,56 @@ export async function attach(
...
+    if (!telemetryReporter) {
+        telemetryReporter = createTelemetryReporter(context);
+    }
+
+    telemetryReporter.sendTelemetryEvent('command/attachToCurrentDebugTarget');
+
+    // Attempt to attach to active CDP target
+    const session = vscode.debug.activeDebugSession;
+    if (!session) {
```
vidorteg, line 271: can we add a test for this scenario? one where the mock value of:
vscode.debug.activeDebugSession = true;

bgoddar: For sure!

vidorteg (approved):
Left a small comment, but this looks great! thanks @bgoddar!

package.json
```diff
@@ -596,7 +597,9 @@
         "download-edge": "node scripts/downloadAndExtractEdge.js"
     },
     "dependencies": {
+        "bufferutil": "4.0.1",
         "puppeteer-core": "9.1.1",
+        "utf-8-validate": "5.0.2",
```
bgoddar, line 602: These were needed for me to be able to create a VSIX. Otherwise it failed in an npm step

src/extension.ts
bgoddar: Done!

package.json
vidorteg: seems to be associated with the websocket so importing them should be fine, although I'm not sure why they were not required before as I don't see any new code bringing them in 😕. Could it be node version?

mliao95:
Took a look at the tests and they look good.  Thanks!

flynnolivia (approved):
Tests LGTM too!

package.json
bgoddar: Yeah, The error message I was seeing was these were marked as 'peer' dependencies of ws, which means I needed to manually install. LMK if I should remove from this PR, happy to do that

leotlee (approved):

package.json
bgoddar: Going to land as is since these are required for running `npm run package` for me. We can follow up if we decide to remove later

PR #397, opened 2021-05-20
Adding a docs folder to allow for smaller README

codepo8:


docs/index.md
```diff
@@ -0,0 +1,129 @@
+
+## Table of Contents
```
tobybaratta, line 2: Question: Many of the links in this TOC don't actually exist anymore. With the README getting too long, will the docs here get split up and cleaned up for accuracy as well? If you're considering fixing those up out of scope for this PR, is there a follow up todo for at minimum cleaning up the links, but preferably updating the docs entirely. I don't want to put in a PR killing the deadlinks and doing some cleanup/clarity just to hit merge conflicts. ;)

Suggestion for follow-up: Consider adding markdown-link-check to your build actions, to make sure that they stay up to date in the future?

codepo8:
Yes, this was a test to have the pr in there. We are cleaning up the readme
and create distinct sections for each feature instead of one long document.

vidorteg (changes requested):
Might be missing something but why is this PR deleting a lot of the ts files in src/ and test?

PR #400, opened 2021-05-27
[Telemetry]: Remove unnecessary telemetry

mliao95:
This PR removes these telemetry events

- `view/list`
    - This fires when `getChildren` is called
    - `getChildren` is called in many different ways, so this isn't super useful.
- `view/refresh`
    - This is called when `refresh()` is called which calls `getChildren`.
    - Not entirely useful for the same reason as `view/list` 
- `command/attach`
    - This only triggers when the browser instance is first attached to.  We already have `command/launch` to track every new browser target.
- `command/attach/list`
    - Same as `command/attach`
- `command/attach/devtools`
    - Same as `command/attach`

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -169,8 +169,6 @@ export async function attach(
     }
 
     const telemetryProps = { viaConfig: `${!!config}`, withTargetUrl: `${!!attachUrl}` };
-    telemetryReporter.sendTelemetryEvent('command/attach', telemetryProps);
-
     const { hostname, port, useHttps, timeout } = getRemoteEndpointSettings(config);
 
     // Get the attach target and keep trying until reaching timeout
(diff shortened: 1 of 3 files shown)


flynnolivia (approved):
LGTM!

src/extension.ts
```diff
@@ -169,8 +169,6 @@ export async function attach(
     }
 
     const telemetryProps = { viaConfig: `${!!config}`, withTargetUrl: `${!!attachUrl}` };
-    telemetryReporter.sendTelemetryEvent('command/attach', telemetryProps);
```
vidorteg, line 172: This might still be useful if the user has already launched the browser (by themselves) and attach to it. IIRC Launch telemetry only covers when we launch the browser.

vidorteg (approved):
LGTM - 1 comment

src/extension.ts
mliao95: Could you double check to see if this fires upon attach to an existing target? It doesn't seem to fire for me.  Currently, the only extension-side telemetry we fire for attaching to an existing target are the button click for the attach button and the websocket connection.

vidorteg: yep you are right not firing either :), so we are good to go.

PR #401, opened 2021-05-28
[Telemetry]: Implement extra launch parameters and websocket session length

mliao95:
This PR adds extra parameters to existing telemetry events.  Here are the changes:

- `command/launch`
    - added extra parameters `browserType` and `isHeadless`
    - `browserType` can be `Default`, `Stable`, `Beta`, `Dev` or `Canary`
    - `isHeadless` can be `true` or `false`
    - Example: 
        - `command/launch: {"viaConfig":"false","browserType":"Default","isHeadless":"false"}`
- `websocket/dispose`
    - added extra parameter `sessionTime` which tracks length the DevTools extension was connected to that target
    - Example:
        - `websocket/dispose: undefined, {"sessionTime":489728.6491000056}`
- `user/buttonPress`
    - We now differentiate whether the `.launch` command is triggered through the plus button or the launch instance blue button.

### DIFF
--- a/package.json
+++ b/package.json
@@ -538,7 +538,7 @@
         "viewsWelcome": [
             {
                 "view": "vscode-edge-devtools-view.targets",
-                "contents": "Launch an instance of Microsoft Edge to begin inspecting and modifying webpages.\n[Launch Instance](command:vscode-edge-devtools-view.launch)",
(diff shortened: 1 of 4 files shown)


flynnolivia (approved):
LGTM. I'm assuming the session duration is in milliseconds, is that right? We should make a note of that in the spreadsheet.

PR #404, opened 2021-06-02
[Telemetry]: Add telemetry for file extension types

mliao95:
This PR adds telemetry that sends the used file extension types in the user's currently active workspace.  The telemetry is sent upon the activation of the extension.

- `workspace/metadata`
    - example:
        - `workspace/metadata: undefined, {"css":1,"js":1,"html":1,"json":2,"total":5}`

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -27,6 +27,7 @@ import {
     SETTINGS_VIEW_NAME,
     getActiveDebugSessionId,
     getJsDebugCDPProxyWebsocketUrl,
+    reportFileExtensionTypes,
     reportUrlType,
 } from './utils';
 import { LaunchConfigManager } from './launchConfigManager';
@@ -160,6 +161,7 @@ export function activate(context: vscode.ExtensionContext): void {
             void vscode.env.openExternal(vscode.Uri.parse('https://github.com/microsoft/vscode-edge-devtools/blob/master/README.md'));
         }));
     void vscode.commands.executeCommand('setContext', 'titleCommandsRegistered', true);
+    void reportFileExtensionTypes(telemetryReporter);
 }
 
 export async function attach(
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -631,18 +631,21 @@ export function reportUrlType(url: string, telemetryReporter: Readonly<Telemetry
 
 export async function reportFileExtensionTypes(telemetryReporter: Readonly<TelemetryReporter>): Promise<void> {
     const files = await vscode.workspace.findFiles('**/*.*', '**/node_modules/**');
-    const extensionSet: Set<string> = new Set();
+    const extensionMap: Map<string, number> = new Map<string, number>();
     for (const file of files) {
         const extension: string | undefined = file.path.split('.').pop();
         if (extension) {
-            extensionSet.add(extension);
+            const currentValue = extensionMap.get(extension);
+            if (currentValue) {
+                extensionMap.set(extension, currentValue + 1);
+            } else {
+                extensionMap.set(extension, 1);
+            }
         }
     }
-    let fileTypes = '';
-    for (const extension of extensionSet.keys()) {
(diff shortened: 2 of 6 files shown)


src/extension.ts
```diff
@@ -160,6 +161,7 @@ export function activate(context: vscode.ExtensionContext): void {
             void vscode.env.openExternal(vscode.Uri.parse('https://github.com/microsoft/vscode-edge-devtools/blob/master/README.md'));
         }));
     void vscode.commands.executeCommand('setContext', 'titleCommandsRegistered', true);
+    void reportFileExtensionTypes(telemetryReporter);
```
mliao95, line 164: Is everyone okay with checking the extension types upon activation?  I didn't want it to fire on launch just so we don't dilute this event.  Open to suggestions

src/utils.ts
```diff
@@ -629,6 +629,22 @@ export function reportUrlType(url: string, telemetryReporter: Readonly<Telemetry
...
+    const extensionSet: Set<string> = new Set();
+    for (const file of files) {
+        const extension: string | undefined = file.path.split('.').pop();
+        if (extension) {
+            extensionSet.add(extension);
+        }
+    }
+    let fileTypes = '';
+    for (const extension of extensionSet.keys()) {
+        fileTypes += extension + ';';
```
mliao95, line 643: Right now, the property value is set as `"css;js;html;json;"` since the property value must be a string. We could search for `".js;"` in the property value in Kusto later to check if a given file extension type exists. If there is a better way to organize this information, I am open to changing this as well.

bgoddar: Hmm, I think a better approach might be using the "measurements" field and using counts or only adding properties we care about.
Something like
``` 
sendTelemetryEvent('workspace/filetypes', undefined, {
   css: 1,
js: 1,
html: 1,
jsx: 0,
...
});
```

bgoddar: I think this might be easier to parse from kusto

bgoddar (changes requested):
Small change request about how we report filetypes

src/extension.ts
flynnolivia: Reporting on activation sounds good to me. I'm not 100% sure but I think if the user opens a new folder, the window will reload and activate the extension again. So if that's the case, we should get reports from multiple projects.
Also, I did a little digging to the textDocuments property of the vscode.workspace earlier, and this seems to be the files that are open in the editor at a given time. It could potentially be useful to see which files they actually have open when they launch the extension (as opposed to the files that make up the entire project), but that should be a separate event I think.

flynnolivia (commented):
Generally lgtm, but I agree with Brandon's idea about using measurements instead of properties and we should confirm that the event is fired again when a new folder is opened

src/utils.ts
mliao95: Hmm yeah I was considering this approach.  Since we don't have an exact list of file extensions we care about, I was thinking that the object would only have existing file extensions.  In that case everything would just be set to 1 which I thought was kind of redundant.  

For example:
`{js: 1, jsx: 1, css: 1}` or `{css: 1, js: 1, json: 1}`

I'm going to run with your idea and maybe change the value to account for number of files with each file extension such as:

`{js: 2, css: 1, json: 1, total: 4}`

bgoddar: Sounds good to me!

src/extension.ts
mliao95: Double checked and opening a new folder will re-launch VS Code and will refire the telemetry event.  However, opening a file will not relaunch nor refire, but I don't think we care about that scenario too much anyway.

src/utils.ts
mliao95: New format: `workspace/metadata: undefined, {"css":1,"js":1,"html":1,"json":2,"total":5}`

bgoddar (approved):
LGTM!

PR #406, opened 2021-06-03
[Telemetry]: Adding extension settings related telemetry

mliao95:
This PR implements two telemetry events related to the extension settings:

- `user/settingsChangedAtLaunch`
    - This event fires upon extension activation and displays all extension settings that differ from the default value
    - Example: `user/settingsChangedAtLaunch: {"enableNetwork":"false","themes":"Light"}, undefined`
- `user/settingsChanged`
    - This event fires whenever a user changes a setting related to our extension
    - Example: `user/settingsChanged: {"enableNetwork":"false"}, undefined`

Going to try something new and do code explanation via comments. Marking them all as resolved so they are collapsible, but feel free to reply to and/or activate them if you have a question regarding that section.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -28,6 +28,8 @@ import {
     getActiveDebugSessionId,
     getJsDebugCDPProxyWebsocketUrl,
     reportFileExtensionTypes,
+    reportChangedExtensionSetting,
+    reportExtensionSettings,
     reportUrlType,
 } from './utils';
 import { LaunchConfigManager } from './launchConfigManager';
@@ -166,6 +168,8 @@ export function activate(context: vscode.ExtensionContext): void {
         }));
     void vscode.commands.executeCommand('setContext', 'titleCommandsRegistered', true);
     void reportFileExtensionTypes(telemetryReporter);
+    reportExtensionSettings(telemetryReporter);
+    vscode.workspace.onDidChangeConfiguration(event => reportChangedExtensionSetting(event, telemetryReporter));
 }
 
 export async function attach(
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -99,25 +99,6 @@ export const SETTINGS_DEFAULT_ATTACH_INTERVAL = 200;
 
 const WIN_APP_DATA = process.env.LOCALAPPDATA || '/';
 const msEdgeBrowserMapping: Map<BrowserFlavor, IBrowserPath> = new Map<BrowserFlavor, IBrowserPath>();
-const extensionSettingsList: string[] = [
-    'hostname',
-    'port',
-    'useHttps',
-    'defaultUrl',
-    'userDataDir',
-    'webRoot',
-    'pathMapping',
-    'sourceMapPathOverrides',
-    'sourceMaps',
-    'autoAttachViaDebuggerForEdge',
-    'enableNetwork',
-    'showWorkers',
-    'whatsNew',
-    'headless',
-    'timeout',
-    'browserFlavor',
-    'themes',
-];
 
 export interface IRemoteTargetJson {
     [index: string]: string;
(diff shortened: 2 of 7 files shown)


src/extension.ts
```diff
@@ -160,6 +162,8 @@ export function activate(context: vscode.ExtensionContext): void {
             void vscode.env.openExternal(vscode.Uri.parse('https://github.com/microsoft/vscode-edge-devtools/blob/master/README.md'));
         }));
     void vscode.commands.executeCommand('setContext', 'titleCommandsRegistered', true);
+    reportExtensionSettings(telemetryReporter);
+    vscode.workspace.onDidChangeConfiguration(event => reportChangedExtensionSetting(event, telemetryReporter));
```
mliao95, line 166: The `vscode.ConfigurationChangeEvent` only contains one function which is `affectsConfiguration(setting)` which returns a bool that checks whether the changed event matches the passed in setting string.

src/utils.ts
```diff
@@ -614,6 +614,58 @@ async function verifyFlavorPath(flavor: BrowserFlavor | undefined, platform: Pla
     }
 }
 
+type ExtensionSettings = [string, boolean | string | {[key: string]: string} | undefined];
+
+export function reportExtensionSettings(telemetryReporter: Readonly<TelemetryReporter>): void {
+    const extensionSettingsList = Object.entries(vscode.workspace.getConfiguration(SETTINGS_STORE_NAME)).splice(4) as Array<ExtensionSettings>;
```
mliao95, line 620: This is the best way I could get a list of all the extension settings.  The `workspaceConfiguration` object has getters, but nothing to return an iterable object.  However, the `workspaceConfiguration` object contains each setting as a property.  I splice it at the 4th index since the first 4 properties are all functions related to the `workspaceConfiguration` class and the rest are the extension properties under `SETTINGS_STORE_NAME`.

`extensionSettingsList` ends up looking something like:
```
[
  ["enableNetwork", "false"],
  ["themes", "Light"],
  ["screencastCursorType", "Default"],
]
...
```

src/utils.ts
```diff
@@ -614,6 +614,58 @@ async function verifyFlavorPath(flavor: BrowserFlavor | undefined, platform: Pla
...
+    const settings = vscode.workspace.getConfiguration(SETTINGS_STORE_NAME);
+    const changedSettingsMap: Map<string, string> = new Map<string, string>();
+    for (const currentSetting of extensionSettingsList) {
+        const settingName: string = currentSetting[0];
+        const settingValue: boolean | string | {[key: string]: string} | undefined = currentSetting[1];
+        const settingInspect = settings.inspect(settingName);
```
mliao95, line 626: `settingInspect` is used to grab the default value for a setting.  We use this to compare to the current value to catch any differences.

src/utils.ts
```diff
@@ -614,6 +614,58 @@ async function verifyFlavorPath(flavor: BrowserFlavor | undefined, platform: Pla
...
+        if (settingInspect) {
+            const defaultValue = settingInspect.defaultValue;
+            if (settingValue !== undefined && settingValue !== defaultValue) {
+                if (defaultValue && typeof defaultValue === 'object' && typeof settingValue === 'object') {
```
mliao95, line 630: `object` value settings are kind of tricky.  For our extension, these are the path mapping ones where the values look something like this and is edited directly in `settings.json`:
```
{
    "/": "${workspaceFolder}"
}
```

When trying to change this setting, it actually links you directly to the `settings.json` file like so:
image

When retrieving these values, it is placed inside a `Proxy` object so we cannot do direct object comparison, but have to compare property by property.

If there is a difference detected between the user entry and the default entry, we add the changed setting to the map and then break the loop.

src/utils.ts
```diff
@@ -614,6 +614,58 @@ async function verifyFlavorPath(flavor: BrowserFlavor | undefined, platform: Pla
...
+                        }
+                    }
+                } else {
+                    changedSettingsMap.set(settingName, settingValue.toString());
+                }
+            }
+        }
+    }
+    const changedSettingsObject = {};
+    Object.assign(changedSettingsObject, ...[...changedSettingsMap.entries()].map(([k, v]) => ({[k]: v})));
```
mliao95, line 644: Converting the map to an object to send with the telemetry event

src/utils.ts
```diff
@@ -614,6 +614,56 @@ async function verifyFlavorPath(flavor: BrowserFlavor | undefined, platform: Pla
...
+    telemetryReporter.sendTelemetryEvent('user/settingsChangedAtLaunch', changedSettingsObject);
+}
+
+export function reportChangedExtensionSetting(event: vscode.ConfigurationChangeEvent, telemetryReporter: Readonly<TelemetryReporter>): void {
```
mliao95, line 648: In this function, we are just iterating over each of the extension's settings to see if the change setting from the event matches any of our settings.  If it does, we send the telemetry event with an object with the format 
`{settingName: currentValue}`

test/utils.test.ts
```diff
@@ -971,8 +972,71 @@ describe("utils", () => {
...
+                                return {defaultValue: undefined};
+                        }
+                    },
+                    fillerPropertyOne: 'Need to add filler values since the extension related telemetry functions removes the first 4 entries to create an array of extension settings',
+                    fillerPropertyTwo: 'The first 4 entries for a workspaceConfiguration object are class related functions, the rest are all extension related settings.',
+                    themes: 'System preference',
+                    whatsNew: 'true',
+                    enableNetwork: 'false',
+                }
+            });
```
flynnolivia, line 1017: I think we can save some repeating lines of code if we make a const with the getConfiguration object from helpers.ts and then use the spread operator. Like this:
```
const getConfig = vscodeMock.workspace.getConfiguration();
vscodeMock.workspace.getConfiguration.mockImplementation(() => {
     return {
          ...getConfig,
          ...other properties...
     }
});
```

test/utils.test.ts
```diff
@@ -971,8 +972,71 @@ describe("utils", () => {
 
             for (let i = 0; i < input.length; i++) {
                 utils.reportUrlType(input[i], reporter);
-                expect(reporter.sendTelemetryEvent).toBeCalledWith('user/browserNavigation', { 'urlType': expected[i] })
+                expect(reporter.sendTelemetryEvent).toBeCalledWith('user/browserNavigation', { 'urlType': expected[i] });
```
flynnolivia, line 975: Weird that this was not caught by the linter during my PR adding this 🤔

flynnolivia (commented):
Lgtm for the most part, but one suggestion

test/utils.test.ts
mliao95: Good suggestion! Adding this change now.

flynnolivia (approved):

Issue #407, opened 2021-06-07
Debugger Integration: CSS Styles may not load if debug target is paused.


bgoddar:
When using the Inspect icon to attach to the current debug target, the Styles pane may show as blank as users select different DOM nodes.

This can occur because enabling the CSS domain isn't possible when the debugger is paused, however, there isn't an indication to users this is the case. This can be fixed by:
1. showing a prompt in the styles pane to unpause/refresh if no styles are available
2. Making a change to CDP/blink to address the issue

More info from original email below:
it seems that the CSS domain can't be enabled if the debugger is currently paused on a breakpoint. What I see is a call to CSS.enable(), then followup calls that result in errors because they don't wait for enablement to happen.

Once the debugger is unpaused, the browser replies to CSS.enable() and styles show up.

bgoddar:
After digging into the limitations in V8 renderer, we have settled on the message for now. Our reasoning is that this would be a fairly significant architectural change to chromium that would be tough to justify given that the inbox devtools cannot hit this particular issue, and it has a fairly trivial workaround. But I'll keep an eye out for feedback

Issue #341, opened 2021-04-07
Devtools should be localized as in the browser.


vidorteg:
Enable localization of the Devtools as in the browser

vidorteg:
Need to also consider the scenario where we are patching with localizable content.  e.g

https://github.com/microsoft/vscode-edge-devtools/pull/410

PR #410, opened 2021-06-08
Patch Styles panel to inform users to resume if debugger paused

bgoddar:
This PR patches the styles pane to inform users that they may need to resume or refresh the target to get styles information. The CSS domain cannot be enabled when the debugger is paused, which leads to a blank styles pane. This updates the default message to inform users what to do. This does not occur for standalone/inbox devtools as the css domain is always enabled during initialization, before any breakpoints can be hit. However, this can be hit fairly easily using the "inspect button" when VSCode's JS Debugger is currently attached to a target that is paused.

Before:
image

After:
image

Addresses #407

### DIFF
--- a/src/host/polyfills/simpleView.ts
+++ b/src/host/polyfills/simpleView.ts
@@ -114,6 +114,15 @@ export function applyStylesToggleFocusPatch(content: string): string | null {
     return replaceInSourceCode(content, pattern, replacementText);
 }
 
+export function applyNoMatchingStylesPatch(content: string): string | null {
+    // Patch to inform user to refresh/resume target to get CSS information when attaching to a paused target.
+    const pattern = /this\._noMatchesElement\.textContent\s*=\s*(ls\s*`No matching selector or style`);/g;
+    const replacementText = `
+       this._noMatchesElement.innerHTML = ls \`No matching selector or style.<br />Styles may not be available if target was paused when opening Edge DevTools.<br />Please resume or refresh the target.\`;
+    `;
+    return replaceInSourceCode(content, pattern, replacementText);
+}
+
 export function applyQuickOpenPatch(content: string): string | null {
     // This patch removes the ability to use the quick open menu (CTRL + P)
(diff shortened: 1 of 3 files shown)


src/host/polyfills/simpleView.ts
```diff
@@ -114,6 +114,15 @@ export function applyStylesToggleFocusPatch(content: string): string | null {
     return replaceInSourceCode(content, pattern, replacementText);
 }
 
+export function applyNoMatchingStylesPatch(content: string): string | null {
+    // Patch to inform user to refresh/resume target to get CSS information when attaching to a paused target.
+    const pattern = /this\._noMatchesElement\.textContent\s*=\s*(ls\s*`No matching selector or style`);/g;
+    const replacementText = `
+       this._noMatchesElement.innerHTML = ls \`No matching selector or style.<br />Styles may not be available if target was paused when opening Edge DevTools.<br />Please resume or refresh the target.\`;
```
bgoddar, line 121: @vidorteg I'm curious what your thoughts are on this particular patch when it comes to Loc. I know this will need to be updated for Loc V2, but I'm also curious how we are going to get updated Loc files for patches when we update strings like this

vidorteg: Thanks for bringing this up. I made a note on the https://github.com/microsoft/vscode-edge-devtools/issues/341 to include this case in the localization task.

vidorteg: should not be a blocker for now as the content meet the bar by being localizable.

vidorteg, line 121: layout should be done outside of localization messages as the library does not support it. So in this case this needs to be split in 3 different parts.

vidorteg (approved):
LGTM % minor change

src/host/polyfills/simpleView.ts
bgoddar: Ahh, good call. Wasn't sure the best way to do this, but that makes alot of sense. I'll go ahead and update

Issue #403, opened 2021-06-01
Hide DevTools Console output when attached to a debug session


connor4312:
This seems to cause some confusion (https://github.com/microsoft/vscode/issues/125071) and is redundant when launching from a js-debug session which already has output in the Debug Console

bgoddar:
Honestly, I don't think it makes much sense for us to include a limited/one-way console when instead we should just direct users to JS debug. I'll update with a Deprecation notice directing users to JS Debug, and we can remove entirely in a future update.

bgoddar:
Hmmm. Actually, I guess the one case it might be useful is attaching to a target we are not using a debug session for. I'll discuss with the team if the best course of action is removing entirely or just disabling when attached to active target

codepo8:
This is an upcoming scenario now.

bgoddar:
@codepo8 this is already checked in. Re-closing

bgoddar:
My mistake. Not checked in yet, but will be checked in today :)

codepo8:
Added in the latest version of the extension

PR #411, opened 2021-06-09
Remove console output in favor of JSDebug complete CDP Console

bgoddar:
This PR removes our console polyfills and replaces our current Devtools Console output with a single message that informs users to use JSDebug instead. Since JSDebug is a complete console experience already built into VSCode, it does not make sense for us to offer a duplicated and subpar one-way console output

Before:
image

After:
image

JSDebug Console:
image

Addresses #403

### DIFF
--- a/postBuildStep.ts
+++ b/postBuildStep.ts
@@ -31,7 +31,6 @@ import {
     applyPersistDrawerTabs,
     applyQueryParamsObjectPatch,
     applyRemoveBreakOnContextMenuItem,
-    applyRerouteConsoleMessagePatch,
     applyContextMenuRevealOption,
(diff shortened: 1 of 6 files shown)


bgoddar:
I'm going to also stand up an alternate PR that just disables when we have attached through JSDebug, in case we want to keep this around for attaching to arbitrary sites. However, I prefer this solution

bgoddar:
> I'm going to also stand up an alternate PR that just disables when we have attached through JSDebug, in case we want to keep this around for attaching to arbitrary sites. However, I prefer this solution

Alternate solution is available here #412

bgoddar:
Closing in favor of #412  for now

PR #412, opened 2021-06-09
Debugger Integration: Only initialize Console when not using JSDebug proxy

bgoddar:
The Devtools Console output channel is redundant when using JSDebug to attach to a target. JSDebug provides an interactive Debug console with support for receiving console messages and evaluating expressions. This changes updates our console output to re-direct users to JSDebug when attached via the CDP share from JSDebug.

Before:
image

After:
image

Addresses #403


bgoddar:
This is an alternate to #411

src/devtoolsPanel.ts
```diff
@@ -54,9 +54,18 @@ export class DevToolsPanel {
...
+            // Direct users to the Debug Console
+            this.consoleOutput.appendLine('// Microsoft Edge Devtools Extension:');
+            this.consoleOutput.appendLine('// You have connected to a target using Visual Studio Code\'s JavaScript Debugger.');
+            this.consoleOutput.appendLine('// Please use the "Debug Console" to view console messages from your webpage and evaluate expressions.');
+        } else {
+            // Provide 1-way console when attached to a target that is not the current debug target
+            this.consoleOutput.appendLine('// Microsoft Edge Devtools Extension:');
+            this.consoleOutput.appendLine('// This Output window displays the DevTools extension\'s console output in text format.');
+            this.consoleOutput.appendLine('// Note that this feature is only unidirectional and cannot communicate back to the DevTools.');
+            this.consoleOutput.appendLine('');
```
bgoddar, line 67: I think it might be worth adding a line here informing users that aren't connected with JSDebug that they could get a complete console experience there. Something like:

"If you would like to evaluate expressions and interact with the JavaScript on your page, please use Visual Studio Code's JavaScript Debugger to attach to your target"

mliao95 (approved):
LGTM, thanks

PR #417, opened 2021-06-14
Transitioning What's New to Welcome tab

mliao95:
This PR reformats and updates the `releaseNote` patches to work with the new Welcome tab introduced in Edge DevTools version 89.  It also contains new release notes that cover the biggest changes since version 1.1.9 and from the DevTools versions 89, 90 and 91.

### DIFF
--- a/src/host/polyfills/simpleView.ts
+++ b/src/host/polyfills/simpleView.ts
@@ -18,7 +18,7 @@ const enum KeepMatchedText {
     AtEnd = 2
 }
 
-const isDrawerEnabled = '(Root.Runtime.vscodeSettings.enableNetwork || Root.Runtime.vscodeSettings.whatsNew)';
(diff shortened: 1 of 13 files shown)


src/host/polyfills/simpleView.ts
```diff
@@ -18,7 +18,7 @@ const enum KeepMatchedText {
     AtEnd = 2
 }
 
-const isDrawerEnabled = '(Root.Runtime.vscodeSettings.enableNetwork || Root.Runtime.vscodeSettings.whatsNew)';
+const isDrawerEnabled = '(Root.Runtime.vscodeSettings.enableNetwork || Root.Runtime.vscodeSettings.welcome)';
```
bgoddar, line 21: nit: don't need drawer with welcome tab AFAIK. I'll fix after merge

PR #418, opened 2021-06-15
Reduce Extension Bundle size

bgoddar:
This PR reduces our extension bundle size in 2 ways:

1. Add a .vscodeignore to the project. Without this file, the vsce package bundles everything inside the project directory (node modules, documentation pictures, source code, etc). This allows us to specify which directories and files are needed at runtime, and exclude everything else.
2. Minify the devtools code using uglify-js. This was added because the update to v91 forces us to use a debug build of devtools for polyfills. This allows us to minify again after applying polyfills

This reduces our vsix size to <8MB when used in conjunction with the update to v91 in #414


.vscodeignore
```diff
@@ -0,0 +1,14 @@
...
+**/*.js.map
+*ignore
+*.vsix
+package-lock.json
+img/
+node_modules/
+out/edge/
+src/
+scripts/
+startpage/
```
mliao95, line 13: We probably need to keep `startpage/index.html` since the extension defaults to accessing this page when launching a target.

mliao95 (commented):
Requesting a small change, but otherwise LGTM

.vscodeignore
bgoddar: Good call. I thought this exclusion would be restricted to top level, but it was also applying to the out/startpage directory. Updated to include out/startpage but exclude startpage/

bgoddar: image
After update

mliao95 (approved):
LGTM

PR #424, opened 2021-06-22
Minification: add ability to skip minification on problematic files

bgoddar:
This PR adds the ability to skip files that don't work with the uglify.js minification module we are using. This addresses an issue in ResourceLoader.js that was preventing proper source map resolution.


mliao95 (approved):
Tested with webpack project and LGTM.

PR #430, opened 2021-07-09
[Linting] Re-enable line endings rule and run autofix

flynnolivia:
I had to disable one rule when migrating to ESLint (requiring LF and not CRLF), but I am re-enabling it now. The change involves:
- Uncommenting the rule in .eslintrc.js
- Adding a step to configure git to check out the branch as LF
- Running ESLint's autofix, which converted two files from LF -> CRLF


flynnolivia:
Looks like I don't need to actually check in the rule to trigger the bug-- I'll take reviewers off until I do more investigation

azure-pipelines.yml
```diff
@@ -73,6 +73,9 @@ jobs:
   pool:
     vmImage: 'windows-latest'
   steps:
+  - script: |
+      git config --global core.autocrlf false
+    displayName: Configuring Git
```
bgoddar, line 78: nit: maybe something like "updating Git LF settings" so we remember what it does when looking at ADO

bgoddar (approved):
LGTM with nit

PR #432, opened 2021-07-14
[Telemetry]: Modifying workspace/metadata telemetry to track only relevant file types

mliao95:
Our previous implementation of the `workspace/metadata` telemetry would send a count of every file extension type in the user's workspace. This created thousands of different properties that needed to be categorized for this telemetry event due to the vast number of file extension types.

To reduce the # of properties for this telemetry event, this PR scopes down to only track the relevant file types.  An example of the new event property would look something like this:
`{"css": 1, "html": 0, "js": 1, "json": 1, "jsx": 1, "mjs": 0, "scss": 0, "totalWorkspaceSize": 10, "ts": 0}`

Note that the `totalWorkspaceSize` property will still track the total # of files, not just the total of relevant file types

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -691,17 +691,25 @@ export async function reportFileExtensionTypes(telemetryReporter: Readonly<Telem
         ['scss', 0],
         ['json', 0],
         ['mjs', 0],
+        ['other', 0],
     ]);
     for (const file of files) {
         const extension: string | undefined = file.path.split('.').pop();
-        if (extension && extensionMap.has(extension)) {
-            const currentValue = extensionMap.get(extension);
-            if (currentValue !== undefined) {
-                extensionMap.set(extension, currentValue + 1);
+        if (extension) {
+            if (extensionMap.has(extension)) {
+                const currentValue = extensionMap.get(extension);
+                if (currentValue !== undefined) {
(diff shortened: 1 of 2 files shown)


flynnolivia (approved):
makes sense, lgtm!

src/utils.ts
```diff
@@ -682,19 +682,26 @@ export function reportUrlType(url: string, telemetryReporter: Readonly<Telemetry
...
-    const extensionMap: Map<string, number> = new Map<string, number>();
+    const extensionMap: Map<string, number> = new Map<string, number>([
+        ['html', 0],
+        ['css', 0],
+        ['js', 0],
+        ['ts', 0],
+        ['jsx', 0],
+        ['scss', 0],
+        ['json', 0],
+        ['mjs', 0],
```
bgoddar, line 693: Nit: should we add an 'other' here? I guess we could calculate that info from totalsize - other files. But, it might make it easier on us if we threw everything else into 'other'.

No worries if it doesn't make sense. Only if we thing that information would be useful

bgoddar (approved):
LGTM with nit

src/utils.ts
mliao95: I'll add the `other` for the sake of clarity and so that the file list is self-contained.

PR #437, opened 2021-07-19
Disabling virtual workspaces.

vidorteg:
Edge Devtools for VSCode does not currently plan to support virtual workspaces. We can always revisit this decision if we find that there is enough user push for it.


flynnolivia (approved):
lgtm!

bgoddar (approved):

antross (approved):

PR #443, opened 2021-08-05
Add new DevTools with Frame support to webview host - Localhost only

bgoddar:
This PR adds a new host bundle that allows the VSCode extension to host compatible devtools from a localhost edge-watch environment. This adds a new host bundle and Webview implementation that supports routing messages between the extension and the DevTools inside an iframe. Not all currently supported features have been migrated, and currently only localhost connections are supported. Future PRs will add the ability to target remote provided DevTools and add remaining functionality before shipping new version of the extension

image


src/utils.ts
```diff
@@ -101,6 +104,10 @@ export const SETTINGS_DEFAULT_ATTACH_INTERVAL = 200;
 const WIN_APP_DATA = process.env.LOCALAPPDATA || '/';
 const msEdgeBrowserMapping: Map<BrowserFlavor, IBrowserPath> = new Map<BrowserFlavor, IBrowserPath>();
 
+/** Build-specified flags. */
```
bgoddar, line 107: hmmm, these are currently not playing nice with our tests. I'll see if I can find a fix

src/devtoolsPanel.ts
```diff
@@ -365,6 +365,36 @@ export class DevToolsPanel {
             `;
     }
 
+    private getCDNHtmlForWebview() {
```
mliao95, line 368: Could we rename this and every other instance of `CDN` to `Cdn` (e.g. `getCdnHtmlForWebview`)? I've messed up these capitalized abbreviations in var/func names enough to know this will come back to haunt me later.

See https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions?redirectedfrom=MSDN

mliao95:
Still looking through it, but I think a simple architecture diagram can help paint a bigger picture of the changes in this PR.  

Something like this:
image

You can steal this diagram to work off of here: https://microsoft.sharepoint.com/:p:/t/Edge/EUL2tWFeek9FggsmuGtNSAUBlmBuaJ6Q3JUb5_B-fbWHTg?e=PQPgRJ&nav=eyJzSWQiOjI5MCwiY0lkIjozMzY2NDEyOTIzfQ

bgoddar:
> Still looking through it, but I think a simple architecture diagram can help paint a bigger picture of the changes in this PR.
> 
> Something like this:
> image
> 
> You can steal this diagram to work off of here: https://microsoft.sharepoint.com/:p:/t/Edge/EUL2tWFeek9FggsmuGtNSAUBlmBuaJ6Q3JUb5_B-fbWHTg?e=PQPgRJ&nav=eyJzSWQiOjI5MCwiY0lkIjozMzY2NDEyOTIzfQ

Sounds good! I'll post that here when I publish the review. Good news is that it isn't too different from what we have now

src/common/webviewEvents.ts
```diff
@@ -19,6 +19,17 @@ export const webviewEventNames: WebviewEvent[] = [
...
+export type FrameToolsEvent = 'sendMessageToBackend' | 'openInNewTab' | 'recordEnumeratedHistogram' |
+'recordPerformanceHistogram' | 'reportError' | 'openInEditor';
+export const FrameToolsEventNames: FrameToolsEvent[] = [
+    'sendMessageToBackend',
+    'openInNewTab',
+    'openInEditor',
+    'recordEnumeratedHistogram',
+    'recordPerformanceHistogram',
+    'reportError',
+];
```
antross, line 31: Fun TS trick to avoid repetition (though I'm fine leaving this as-is for consistency with other existing types):

```ts
export const FrameToolsEventNames = [
    'sendMessageToBackend',
    'openInNewTab',
    'openInEditor',
    'recordEnumeratedHistogram',
    'recordPerformanceHistogram',
    'reportError',
] as const;
export type FrameToolsEvent = typeof FrameToolsEventNames[number];
```

src/devtoolsPanel.ts
```diff
@@ -365,6 +365,36 @@ export class DevToolsPanel {
...
+        // the added fields for "Content-Security-Policy" allow resource loading for other file types
+        return `
+            <!doctype html>
+            <html>
+            <head>
+                <meta http-equiv="content-type" content="text/html; charset=utf-8">
+                <meta name="referrer" content="no-referrer">
+                <link href="${stylesUri}" rel="stylesheet"/>
+                <script src="${hostUri}"></script>
+            </head>
```
antross, line 390: We'll want to figure out what the right CSP is to put back here. Ideally we're only allowing the minimum needed capabilities to load and communicate with the frame.

src/host_beta/messageRouter.ts
```diff
@@ -0,0 +1,242 @@
...
+
+    getVscodeSettings(callback: (arg0: Record<string, unknown>) => void): void {
+        const id = this.getHostCallbacksNextId++;
+        this.getHostCallbacks.set(id, callback);
+        encodeMessageForChannel(msg => vscode.postMessage(msg, '*'), 'getVscodeSettings', {id});
+    }
+
+    sendToVscodeOutput(consoleMessage: string): void {
+        encodeMessageForChannel(msg => vscode.postMessage(msg, '*'), 'consoleOutput', {consoleMessage});
+    }
```
antross, line 134: I think these two are obsolete in this context and can be removed.

src/host_beta/messageRouter.ts
```diff
@@ -0,0 +1,242 @@
...
+        // Load the preference via the extension workspaceState
+        const id = this.getHostCallbacksNextId++;
+        this.getHostCallbacks.set(id, callback);
+        encodeMessageForChannel(msg => vscode.postMessage(msg, '*'), 'getState', { id });
+    }
+
+    setPreference(name: string, value: string): void {
+        // Save the preference via the extension workspaceState
+        encodeMessageForChannel(msg => vscode.postMessage(msg, '*'), 'setState', { name, value });
+    }
```
antross, line 81: I think these two are also obsolete in this context and can be removed.

src/host_beta/messageRouter.ts
```diff
@@ -0,0 +1,242 @@
...
+                this.recordPerformanceHistogram(histogramName, duration);
+                return true;
+            case 'reportError':
+                const [type, message, stack, filename, sourceUrl, lineno, colno] = args;
+                this.reportError(type, message, stack, filename, sourceUrl, lineno, colno);
+                return true;
+            case 'sendMessageToBackend':
+                const [cdpMessage] = args;
+                this.sendMessageToBackend(cdpMessage);
+                return true;
```
antross, line 168: Nit: given we're not really validating types here anyway, I'd prefer to reduce the duplication of method names / case statements here. Maybe have the switch cases invoke `encodeMessageForChannel` directly?

src/host_beta/messageRouter.ts
```diff
@@ -0,0 +1,242 @@
...
+        }
+    }
+
+    onMessageFromChannel(e: WebviewEvent, args: string): boolean {
+        switch (e) {
+            case 'getState': {
+                const { id, preferences } = JSON.parse(args) as {id: number, preferences: Record<string, unknown>};
+                this.fireGetHostCallback(id, preferences);
+                break;
+            }
```
antross, line 181: I think this (and the referenced `fireGetHostCallback`) are obsolete as well.

antross (commented):
Works great! Left some comments on additional cleanup I think we can do in the new `host_beta` folder.

src/devtoolsPanel.ts
```diff
@@ -365,6 +365,44 @@ export class DevToolsPanel {
...
+                    content="default-src;
+                    img-src 'self' data: ${this.panel.webview.cspSource};
+                    style-src 'self' 'unsafe-inline' ${this.panel.webview.cspSource};
+                    script-src 'self' 'unsafe-eval' ${this.panel.webview.cspSource};
+                    frame-src 'self' ${this.panel.webview.cspSource} ${cdnBaseUrl};
+                    connect-src 'self' data: ${this.panel.webview.cspSource};
+                ">
+            </head>
+            <body>
+                <iframe id="host" frameBorder="0" src="${cdnBaseUrl}?experiments=true&edgeThemes=true"></iframe>
```
vidorteg, line 400: nit: I suggest to change the id of the iframe as the real host would be the window created by the webview and could cause confusion

src/host_beta/messageRouter.ts
bgoddar: Latest version of the file has incorporated this change, but I'm not 100% sure I like the look better. can folks take a look and lmk what you think?

src/common/webviewEvents.ts
bgoddar: Or do it for all of the files now...

src/host_beta/messageRouter.ts
```diff
@@ -0,0 +1,169 @@
...
+                this.onMessageFromFrame(messageEvent.data.method as FrameToolsEvent, messageEvent.data.args as any[]);
+            } else if (this.toolsFrameWindow) {
+                // Send message from Extension to DevTools
+                parseMessageFromChannel(
+                    messageEvent.data,
+                    extensionMessageCallback,
+                );
+                messageEvent.preventDefault();
+                messageEvent.stopImmediatePropagation();
+                return false;
```
vidorteg, line 54: not sure if this return statement is needed

src/host_beta/messageRouter.ts
antross: Yeah, it reads a bit dense. We could flip this around to keep the methods, drop the switch, and use square-bracket notation to pick the method to dispatch to using `.apply` to pass the args. If we go that route I'd prefer to have an object/instance with *only* those methods so we're not exposing access to anything else.

```ts
const messageHandlers = {
  openInEditor(url: string, line: number, column: number, ignoreTabChanges: boolean): void {
    // Forward the data to the extension
    const request: IOpenEditorData = { column, line, url, ignoreTabChanges };
    encodeMessageForChannel(msg => vscode.postMessage(msg, '*'), 'openInEditor', request);
  }
  // ...
}

onMessageFromFrame(e: FrameToolsEvent, args: any[]): boolean {
  const handler = messageHandlers[e];
  if (handler) {
    handler.apply(messageHandlers, args);
    return true;
  }
  return false;
}
```

antross: But this was a nit to begin with so feel free to go back to what you had before if you'd prefer ;)

src/devtoolsPanel.ts
```diff
@@ -365,6 +365,44 @@ export class DevToolsPanel {
...
+        }
+        const hostPath = vscode.Uri.file(path.join(this.extensionPath, 'out', 'host_beta', 'host.bundle.js'));
+        const hostUri = this.panel.webview.asWebviewUri(hostPath);
+
+        const stylesPath = vscode.Uri.file(path.join(this.extensionPath, 'out', 'common', 'styles.css'));
+        const stylesUri = this.panel.webview.asWebviewUri(stylesPath);
+
+        // the added fields for "Content-Security-Policy" allow resource loading for other file types
+        return `
+            <!doctype html>
```
vidorteg, line 383: we should add the styles to render the frame correctly, it is really small ATM
image

vidorteg (changes requested):
Really nice work @bgoddar , left a couple of minor comments

src/host_beta/mainHost.ts
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { MessageRouter } from './messageRouter';
+export let messageRouter = new MessageRouter(window as unknown as Window);
```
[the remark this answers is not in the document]
bgoddar: ahh, good catches. Left over from the legacy code that is no longer needed.

src/host_beta/messageRouter.ts
```diff
@@ -0,0 +1,169 @@
...
+ * Both the DevTools iframe and the Extension (which owns the ws connection to the browser) will
+ * post messages to the Webview window for communication between the two. This class routes message
+ * to the correct location based on the origin and type of message posted.
+ *
+ */
+export class MessageRouter {
+    private toolsFrameWindow: Window | undefined;
+
+    constructor(webviewWindow: Window | null) {
+        if (!webviewWindow) {
```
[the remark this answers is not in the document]
bgoddar: Now with the way things are set up, this in theory should never hit. I agree though that we should consider some sort of error reporting mechanism from inside the WebView, but I'd like to separate that from this PR

src/host_beta/messageRouter.ts
```diff
@@ -0,0 +1,169 @@
...
+const vscode = acquireVsCodeApi();
+
+/**
+ * Both the DevTools iframe and the Extension (which owns the ws connection to the browser) will
+ * post messages to the Webview window for communication between the two. This class routes message
+ * to the correct location based on the origin and type of message posted.
+ *
+ */
+export class MessageRouter {
+    private toolsFrameWindow: Window | undefined;
```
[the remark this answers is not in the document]
bgoddar: This is to make TS compiler happy. We don't have a reference to toolsFrameWindow at construction, only after DOMContentLoaded, so we have to mark this as possibly undefined. To your other comment above, we should consider adding an error reporting mechanism to allow this to be shown to the user if this ever fails to properly load, but I'd like that to be done separately from this, as we have parity with the legacy system in this approach

src/host_beta/messageRouter.ts
bgoddar: Good call. Not sure why this was present in the legacy code, but I don't think return values in event handlers are ever used

src/devtoolsPanel.ts
bgoddar: So, I used to see this, but I no longer see this particular problem. If you continue to repro, let's call and see if we can figure out what is going on

src/host_beta/messageRouter.ts
bgoddar: Reverted back to this version, as I'm inclined to leave as is. IMO it makes it easier to understand the differences and similarities between the legacy system and this one, as both end up with these functions for encoding messages. I'm open to revisiting this though if folks have strong objections

src/host_beta/messageRouter.ts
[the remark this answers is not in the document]
bgoddar: made a slight change to accept null/undefined to prevent the need for turning null results from the getElementById call to undefined.

I agree this is not ideal and we should add error handling if the dt-window fails to load in a separate PR.

src/host_beta/messageRouter.ts
```diff
@@ -0,0 +1,169 @@
...
+        const request: IOpenEditorData = { column, line, url, ignoreTabChanges };
+        encodeMessageForChannel(msg => vscode.postMessage(msg, '*'), 'openInEditor', request);
+    }
+
+    private openInNewTab(url: string): void {
+        encodeMessageForChannel(msg => vscode.postMessage(msg, '*'), 'openUrl', {url});
+    }
+
+    private sendMessageToBackend(message: string): void {
+        encodeMessageForChannel(msg => vscode.postMessage(msg, '*'), 'websocket', { message });
```
vidorteg, line 148: if we already have a hold for the webviewWindow (passed in the constructor) should we be able to use that instead of using the acquireVSCodeApi?

bgoddar: No, these messages need to be passed to the main extension, not the webview window

src/devtoolsPanel.ts
vidorteg: somehow it stopped after updating VSCode, probably has some kind of caching but it is now fixed :), thanks Brandon

vidorteg (approved):
I dont have any more feedback and code seems solid to me, so approving on my side.

antross (approved):

Issue #413, opened 2021-06-09
Unable to use the extension with VS Code remote


aaronpowell:
_This is a companion to https://github.com/microsoft/vscode-js-debug/issues/1024, but I thought it best to have here as well for tracking._

**Issue**

When using the new inspector feature of `js-debug` to add an Edge inspector to an existing browser, it fails in remote environments as it doesn't detect the extension is installed. This would be due to [`vscode.extensions.all` not working as expected in remotes](https://github.com/microsoft/vscode/issues/82633).

To test a workaround, I created a custom build of `js-debug`, but while it is able to show the inspector panel as expected, the inspector doesn't connect to Edge in a way that you can see the DOM or Network information (the JS console is connected though, so anything written with `console.*` by the browser appears in VS Code).

You can see the issue in action in the following video:

https://user-images.githubusercontent.com/434140/121441596-7696b900-c9cd-11eb-8522-af5d3f205d8b.mp4

It's worth noting that you can use the Edge extension to launch a remotely controlled version of Edge and use the DOM/network inspectors from a remote, so it's likely something to do with the integration between the Edge extension and `js-debug`.

codepo8:
Thank you, this is an interesting problem, we'll add it to the current bug sprint.

aaronpowell:
Make sure you are across the discussion on https://github.com/microsoft/vscode-js-debug/issues/1024 as there's some more insights into it there 😁

vidorteg:
Doing some cleaning. I tested the reported scenario and seems it is working with the latest version of the extension and I also see work done to fix this. Seems the only remaining work was to close this item, please feel free to reopen if this issue is still present

PR #444, opened 2021-08-06
Add "forwardToUi" param when requesting jsdebug proxy connection

bgoddar:
Currently, Edge devtools is unable to use the JsDebug integration when jsdbug is in a remote workspace. This PR adds a new parameter to the request for a CDP proxy connection to indicate that JsDebug should forward CDP calls to the local UI instance of the Devtools extension.

Addresses #413

More information:
https://github.com/microsoft/vscode-js-debug/issues/1024
https://github.com/microsoft/vscode-js-debug/commit/7afc5b2bb68029296d94d2eb79c0852820e61f10

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -280,6 +280,7 @@ export function getActiveDebugSessionId(): string|undefined {
 export async function getJsDebugCDPProxyWebsocketUrl(debugSessionId: string): Promise<string|Error|undefined> {
     try {
         // TODO: update to query location when workspace support added
+        // https://github.com/microsoft/vscode-edge-devtools/issues/383
         const forwardToUi = true;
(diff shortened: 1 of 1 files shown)


src/utils.ts
```diff
@@ -279,9 +279,12 @@ export function getActiveDebugSessionId(): string|undefined {
  */
 export async function getJsDebugCDPProxyWebsocketUrl(debugSessionId: string): Promise<string|Error|undefined> {
     try {
+        // TODO: update to query location when workspace support added
```
bgoddar, line 282: I'm digging up the issue number that we have for this

flynnolivia (approved):
I haven't gotten a chance to test it yet, but generally lgtm based on what I saw in the issue on the js-debug repo and since Connor said it worked

PR #449, opened 2021-08-09
Remote CDN connection logic for devtools

bgoddar:
This PR adds the ability for the extension to choose the legacy (inbox) or CDN based connection to a debug target based on it's version.

It adds a new versionDetectionSocket that sends a single message: CDP's browser.getVersion in order to determine which version of the tools to load, then informs the devtoolsPanel and disposes itself

### DIFF
--- a/src/devtoolsPanel.ts
+++ b/src/devtoolsPanel.ts
@@ -394,6 +394,8 @@ export class DevToolsPanel {
     }
 
     private getCdnHtmlForWebview() {
+        // Default to config provided base uri
+        const cdnBaseUri = this.config.devtoolsBaseUri || this.devtoolsBaseUri;
         const hostPath = vscode.Uri.file(path.join(this.extensionPath, 'out', 'host_beta', 'host.bundle.js'));
         const hostUri = this.panel.webview.asWebviewUri(hostPath);
 
@@ -416,12 +418,12 @@ export class DevToolsPanel {
                     img-src 'self' data: ${this.panel.webview.cspSource};
                     style-src 'self' 'unsafe-inline' ${this.panel.webview.cspSource};
                     script-src 'self' 'unsafe-eval' ${this.panel.webview.cspSource};
-                    frame-src 'self' ${this.panel.webview.cspSource} ${this.devtoolsBaseUri};
+                    frame-src 'self' ${this.panel.webview.cspSource} ${cdnBaseUri};
                     connect-src 'self' data: ${this.panel.webview.cspSource};
                 ">
             </head>
             <body>
-                <iframe id="devtools-frame" frameBorder="0" src="${this.devtoolsBaseUri}?experiments=true&theme=${theme}"></iframe>
+                <iframe id="devtools-frame" frameBorder="0" src="${cdnBaseUri}?experiments=true&theme=${theme}"></iframe>
             </body>
             </html>
             `;
--- a/src/devtoolsPanel.ts
+++ b/src/devtoolsPanel.ts
@@ -41,6 +41,7 @@ export class DevToolsPanel {
     private versionDetectionSocket: BrowserVersionDetectionSocket;
     private consoleOutput: vscode.OutputChannel;
     private timeStart: number | null;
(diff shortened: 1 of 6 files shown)


src/versionSocketConnection.ts
```diff
@@ -0,0 +1,91 @@
...
+        // Dispose socket after version is determined
+        this.dispose();
+        return;
+    }
+
+    private calcBrowserRevision(data: BrowserVersionCdpResponse): string {
+        if (data.id !== 0 || !data.result || !data.result.product && !data.result.revision) {
+            return '';
+        }
+        // product in the form Edg/Major.0.Minor.0 or HeadlessEdg/Major.0.Minor.0
```
bgoddar, line 78: I'm not 100% sure this is true. Anyone know if this is accurate?

src/versionSocketConnection.ts
```diff
@@ -0,0 +1,91 @@
...
+    revision: string;
+}
+
+const MIN_SUPPORTED_BROWSER: BrowserVersion = {
+    major: 94,
+    minor: 975,
+    revision : '',
+};
+
+export class BrowserVerisonDetectionSocket extends EventEmitter {
```
antross, line 30: Minor typo
```suggestion
export class BrowserVersionDetectionSocket extends EventEmitter {
```

src/versionSocketConnection.ts
antross: Poking around I've at least seen the last number not be zero. Probably best to err on the safe side and assume all can be non-zero (assuming you don't actually need to consume `major` or `minor` from other parts of the code).

```ts
const minSupported = '94.0.975.0'.split('.');
// ...
return parts.some((v, i) => v < minSupported[i]) ? '' : currVersion.revision;
```

bgoddar: Updated

antross: Well, perhaps don't use my incorrect code, but the idea stands 😀

src/versionSocketConnection.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export interface BrowserVersionCdpResponse {
+   id: number;
+   result?: {
+       product?: string;
+       revision?: string;
+   }
+}
+
+const MIN_SUPPORTED_VERSION = '94.0.975.0';
```
bgoddar, line 18: TODO: update to actual min version

flynnolivia (commented):
This lgtm. After testing it and going through the changes, nothing looks like an issue to me. I think this should be good once you add the new tests!

antross (approved):

src/versionSocketConnection.ts
bgoddar: Going to leave as is for now to facilitate testing, but this should be updated to the first canary after the last polyfill migrations land

src/devtoolsPanel.ts
```diff
@@ -414,6 +426,17 @@ export class DevToolsPanel {
             `;
     }
 
+    private setBrowserRevision(revision: string) {
+        if (revision !== '') {
+            this.config.isCdnHostedTools = true;
```
mliao95, line 431: NIT: If the devtoolsBaseUri is set in the config, can we check for it and avoid setting up the BrowserVersionDetectionSocket / the version detection logic altogether?

mliao95 (approved):
LGTM with one comment

src/devtoolsPanel.ts
bgoddar: Good call. I'll work on this in the morning. I think I might make separate variables for some of these and make config readonly to make sure we don't override things unintentionally.

bgoddar: Had a quick offline discussion with Michael. I am going to create separate variables to prevent writing the config object within devtools panel. However, I'm not going to skip the version detection logic as it is needed in the follow up headless pr for now. Since this bypass would only apply to local debugging, and could potentially cause confusion when testing headless settings, etc, we thought it would be best to not do that part.

PR #455, opened 2021-08-12
Add theme detection for supported Devtools themes in frame hosted tools

bgoddar:
This PR adds support for setting the devtools extension theme if the user switches to a theme that the devtools frontend supports

### DIFF
--- a/src/common/settingsProvider.ts
+++ b/src/common/settingsProvider.ts
@@ -4,6 +4,25 @@ import * as vscode from 'vscode';
 import { SETTINGS_STORE_NAME } from '../utils';
 import { ThemeString } from './webviewEvents';
 
+const SUPPORTED_THEMES = new Map<string, string>([
+  ['Default Light+', 'light'],
+  ['Visual Studio Light', 'light'],
+  ['Default Dark+', 'dark'],
+  ['Visual Studio Dark', 'dark'],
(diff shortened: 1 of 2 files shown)


src/common/settingsProvider.ts
```diff
@@ -20,6 +39,12 @@ export class SettingsProvider {
     return themeString;
   }
 
+  getNewThemeSettings(): string {
+      const themeSetting = vscode.workspace.getConfiguration().get('workbench.colorTheme');
+      const legacySetting = vscode.workspace.getConfiguration(SETTINGS_STORE_NAME).get('themes');
```
hkal, line 44: What's the difference between these two settings?

hkal (approved):
I like the approach. LGTM.

src/common/settingsProvider.ts
bgoddar: Good question. We currently have a setting specific to the devtools that just selects between light, dark, and system preference. The new function attempts to auto match the current theme, but then falls back to that setting. I plan on updating the setting to include all the new supported themes (or remove it entirely) once the migration is out in stable edge

bgoddar: Tried to make this more clear in updated comments

mliao95 (approved):
LGTM! Thanks for setting this up.

PR #456, opened 2021-08-14
Adding required changes to implement getVsCodeSettings

vidorteg:
Implementing getVsCode Settings for extension with CDN support.
After this PR CDN devtoolswill be able to query for settings in VS Code. I added only two settings by default that seem relevant to the CDN implementation but it can be easily tweaked.


bgoddar:
@vidorteg can we close this PR since we went with a different approach?

PR #458, opened 2021-08-16
Add headless query param support to Cdn DevTools

bgoddar:
This PR updates the browser revision calculation socket to also pass a headless query param to allow the DevTools to know if it is attached to a headless target.

### DIFF
--- a/src/devtoolsPanel.ts
+++ b/src/devtoolsPanel.ts
@@ -42,6 +42,7 @@ export class DevToolsPanel {
     private consoleOutput: vscode.OutputChannel;
     private timeStart: number | null;
     private devtoolsBaseUri: string | null;
+    private isHeadless: boolean;
 
     private constructor(
         panel: vscode.WebviewPanel,
@@ -57,6 +58,7 @@ export class DevToolsPanel {
         this.config = config;
         this.timeStart = null;
         this.devtoolsBaseUri = this.config.devtoolsBaseUri || null;
+        this.isHeadless = false;
         this.consoleOutput = vscode.window.createOutputChannel('DevTools Console');
         if (config.isJsDebugProxiedCDPConnection) {
             // Direct users to the Debug Console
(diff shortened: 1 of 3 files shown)


flynnolivia (approved):
LGTM!

src/devtoolsPanel.ts
```diff
@@ -420,19 +422,20 @@ export class DevToolsPanel {
...
             `;
     }
 
-    private setBrowserRevision(revision: string) {
-        if (revision !== '') {
+    private setBrowserRevision(msg: {revision: string, isHeadless: boolean}) {
+        if (msg.revision !== '') {
             this.config.isCdnHostedTools = true;
-            this.config.devtoolsBaseUri = this.config.devtoolsBaseUri || `https://devtools.azureedge.net/serve_file/${revision}/vscode_app.html`;
+            this.config.devtoolsBaseUri = this.config.devtoolsBaseUri || `https://devtools.azureedge.net/serve_file/${msg.revision}/vscode_app.html`;
```
mliao95, line 434: NIT: If the `devtoolsBaseUri` is set in the config, can we check for it and avoid setting up the `BrowserVersionDetectionSocket` / the version detection logic altogether?

mliao95 (approved):
LGTM with small comment

src/devtoolsPanel.ts
```diff
@@ -420,19 +422,20 @@ export class DevToolsPanel {
...
-                <iframe id="devtools-frame" frameBorder="0" src="${cdnBaseUrl}?experiments=true&theme=${theme}"></iframe>
+                <iframe id="devtools-frame" frameBorder="0" src="${cdnBaseUrl}?experiments=true&theme=${theme}&headless=${this.isHeadless}"></iframe>
             </body>
             </html>
             `;
     }
 
-    private setBrowserRevision(revision: string) {
-        if (revision !== '') {
+    private setBrowserRevision(msg: {revision: string, isHeadless: boolean}) {
```
mliao95, line 431: My small NIT is that the function/event name `setBrowserRevision` doesn't relate to setting and grabbing the headless state.  Since this is a temporary solution to grab the headless state, I'm more okay with leaving it in, but it is a minor change either way.

mliao95 (approved):

src/devtoolsPanel.ts
bgoddar: Thats a good point. I think it makes sense to update the name of the event in response.

PR #459, opened 2021-08-16
Add links to documentation/file an issue to sidebar

flynnolivia:
This PR adds a new section to the extension sidebar, which contains links to documentation and the 'New Issue' page in our repo. It also updates the url in the viewDocumentation command from this repo's README to the official Microsoft documentation.

A few challenges posed by the VSCode Extension API
- [By convention, any command links that are on a line by themselves are displayed as a button](https://code.visualstudio.com/api/references/contribution-points#contributes.viewsWelcome), which prevents us from showing these as links instead of buttons
- The initial height of the tree view sections can't be specified (afaik), so the two sections will take up equal amounts of the sidebar the first time the user opens it, which will show a lot of empty space on the bottom section. Afterwards though, VSCode will remember whatever height they set it to.

Open questions:
- How much deviation from the mock up are we okay with?
- What should the title for the new section be?

Before
sidebar-links-before
After
sidebar-links-after
Mockup:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -577,7 +577,15 @@
             },
             {
                 "view": "vscode-edge-devtools-view.help-links",
-                "contents": "[Documentation](command:vscode-edge-devtools-view.viewDocumentation)\n[File an Issue](https://github.com/microsoft/vscode-edge-devtools/issues/new)"
+                "contents": "[Documentation](command:vscode-edge-devtools-view.viewDocumentation)"
+            },
+            {
+                "view": "vscode-edge-devtools-view.help-links",
+                "contents": "[Report a Bug](https://github.com/microsoft/vscode-edge-devtools/issues/new?template=bug_report.md)"
+            },
+            {
+                "view": "vscode-edge-devtools-view.help-links",
+                "contents": "[Request a Feature](https://github.com/microsoft/vscode-edge-devtools/issues/new?template=feature_request.md)"
             }
         ]
     },
--- a/package.json
+++ b/package.json
@@ -542,6 +542,10 @@
                 {
                     "id": "vscode-edge-devtools-view.targets",
                     "name": "Targets"
+                },
+                {
+                    "id": "vscode-edge-devtools-view.help-links",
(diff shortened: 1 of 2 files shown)


codepo8:
1) If they can't be links, buttons also work - they are even more inviting
to interact.
2) Initial sizing is also not a show stopper.
3) "helpful links" is also a good title.

package.json
```diff
@@ -572,8 +576,8 @@
                 "when": "launchJsonStatus == Supported && isWorkspaceTrusted"
             },
             {
-                "view": "vscode-edge-devtools-view.targets",
-                "contents": "If you have any questions or want to learn more, check the [docs on GitHub](command:vscode-edge-devtools-view.viewDocumentation)."
+                "view": "vscode-edge-devtools-view.help-links",
+                "contents": "[Documentation](command:vscode-edge-devtools-view.viewDocumentation)\n[File an Issue](https://github.com/microsoft/vscode-edge-devtools/issues/new)"
```
bgoddar, line 580: nit: wdyt about creating a github issue template for filing an issue. We might be able to pre-populate with some useful info like extension version, etc?

I'm thinking we could start with something basic like asking for repro steps for a bug, etc

bgoddar (approved):
LGTM with nit about creating an issue template

codepo8:
Excellent idea. Interestingly enough the webhint extension has a great one
but they found out often overwhelms people, so let's keep it light.

package.json
flynnolivia: Yeah that's a great idea! I'll look into making templates on github

codepo8: I just chatted with Rachel, who did the templates for the webhint extension, and she basically said to use the [online editor of GitHub](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository). You can determine if an issue came from the extension or GitHub, but she didn't know if you can automatically include version numbers and such.

PR #462, opened 2021-08-18
[Feature]: Implement CSS mirror editing

mliao95:
This PR adds the cssMirrorContent function for VSCode hosted DevTools that allows mirror editing for the extension when the modified stylesheet is in the VSCode workspace.

In this PR, we set up the necessary infrastructure to consume the `cssMirrorContent` message from the DevTools (similar to `openinEditor`).  This includes adding a webview event, framesocket event, and the right listeners.

Additionally, this PR separates out the code to parse a URL into a URI which originally lives just in the `onSocketOpenInEditor` function.  This is now used in the new `onSocketCssMirrorContent` function which is similar to `onSocketOpenInEditor`, but instead of revealing, overwrites the file using `vscode.workspace.fs.writeFile(uri, encodedNewContent);`.


src/devtoolsPanel.ts
```diff
@@ -302,6 +304,44 @@ export class DevToolsPanel {
...
+        // Parse message and open the requested file
+        const { url, newContent } = JSON.parse(message) as ICssMirrorContentData;
+
+        const uri = await this.parseUrlToUri(url);
+
+        // Finally open and edit the document if it exists
+        if (uri) {
+            const enc = new TextEncoder();
+            const encodedNewContent = enc.encode(newContent);
+            void vscode.workspace.fs.writeFile(uri, encodedNewContent);
```
antross, line 338: Is there an option to replace the contents in the editor without writing the file to disk? Testing this with the live preview extension, I can't complete any CSS edits because when the file saves after my first keystroke the page reloads.

package.json
```diff
@@ -261,6 +273,11 @@
                         "Microsoft Edge Tools for VS Code will use the Edge DevTools light theme",
                         "Microsoft Edge Tools for VS Code will use the Edge DevTools dark theme"
                     ]
+                },
+                "vscode-edge-devtools.mirrorEdits": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Enable edits in the styles pane to be automaticallly applied to the corresponding stylesheet in the workspace. (Only works when targeting Edge Browser versions 94.0.988.0+)"
```
vidorteg, line 280: nit: automatically

postBuildStep.ts
```diff
@@ -121,7 +121,7 @@ async function patchFilesForWebView(toolsOutDir: string) {
     ]);
     // Post built files swap upstream and downstream names
     await patchFileForWebViewWrapper('panels/elements/StylesSidebarPane_edge.js', toolsOutDir, [
-        applyNoMatchingStylesPatch
+        applyNoMatchingStylesPatch,
```
vidorteg, line 124: this seems like an unrelated change

src/devtoolsPanel.ts
```diff
@@ -304,6 +306,61 @@ export class DevToolsPanel {
...
+                errorCode: ErrorCodes.Error,
+                title: 'Error while opening file in editor.',
+                message: e,
+            });
+        }
+    }
+
+    private getDocumentFullRange(textEditor: vscode.TextEditor): vscode.Range {
+        const firstLine = textEditor.document.lineAt(0);
+        const lastLine = textEditor.document.lineAt(textEditor.document.lineCount - 1);
```
vidorteg, line 358: is there any case where this might end up calling lineAt(-1)? e.g if the user deletes the whole content of the file and then reopens it? (not sure if the scenario is possible just thought about this corner case)

src/devtoolsPanel.ts
```diff
@@ -467,9 +504,11 @@ export class DevToolsPanel {
     private setCdnParameters(msg: {revision: string, isHeadless: boolean}) {
         if (msg.revision !== '') {
             this.config.isCdnHostedTools = true;
+            void vscode.commands.executeCommand('setContext', 'isCdnHostedTools', true);
```
vidorteg, line 507: I think this should be awaited

src/host_beta/messageRouter.ts
```diff
@@ -70,6 +71,10 @@ export class MessageRouter {
                 const [url, line, column, ignoreTabChanges] = args;
                 this.openInEditor(url, line, column, ignoreTabChanges);
                 return true;
+            case 'cssMirrorContent':
+                const [cssUrl = url, newContent] = args;
```
vidorteg, line 75: why is `url` needed here?

test/utils.test.ts
```diff
@@ -1014,8 +1014,6 @@ describe("utils", () => {
             vscodeMock.workspace.getConfiguration.mockImplementation(() => {
                 return {
                     ...originalWorkspaceMockConfig,
-                    fillerPropertyOne: 'Need to add filler values since the extension related telemetry functions removes the first 4 entries to create an array of extension settings',
-                    fillerPropertyTwo: 'The first 4 entries for a workspaceConfiguration object are class related functions, the rest are all extension related settings.',
```
vidorteg, line 1018: are these changes required for this PR

vidorteg (changes requested):
Overall looks great, there are a couple of comments that might be good to address before checking in.

test/utils.test.ts
mliao95: These actually somewhat related to make the [reportExtensionSettings test](https://github.com/microsoft/vscode-edge-devtools/blob/52d9d1896f28f78b6257740c33383e5324e93024/test/utils.test.ts#L1027) work after adding `update` [calls](https://github.com/microsoft/vscode-edge-devtools/blob/aac3c68f4b7cd343ed0385b66fa9680848f2acfd/src/extension.ts#L182) in `extension.ts`

Kind of convoluted, but here is the explanation if you care to read:

Basically  `reportExtensionSettings` is a function that grabs all settings and their values from the workspace config object via `Object.entries(vscode.workspace.getConfiguration(SETTINGS_STORE_NAME))`.  The first four entries in the array generated by `Object.entries` are the methods that the config object has which we aren't interested in. In the `reportExtensionSettings` function, [we splice the first four indices away](https://github.com/microsoft/vscode-edge-devtools/blob/main/src/utils.ts#L682) and the remaining array contains all the settings information.

I initially added these filler properties since we only [mocked 2](https://github.com/microsoft/vscode-edge-devtools/pull/462/files#diff-1ba4fa139870853d3020cb0394b69c4f19e55cd6da45d8442bff47e17e7dee94L82-R111) of the 4 methods of `getConfiguration`. Since the `reportExtensionSettings` function spliced away 4 indices, in the test environment, it would splice away 2 of the mocked functions and then 2 settings objects.  I fixed this by adding filler properties to the `getConfiguration` call so that it would splice the 2 mocked methods and the 2 filler properties, leaving all the settings objects intact.

I needed to mock `config.update` after adding `update` [calls](https://github.com/microsoft/vscode-edge-devtools/blob/aac3c68f4b7cd343ed0385b66fa9680848f2acfd/src/extension.ts#L182) in `extension.ts` and realized the obvious solution would be to mock the remaining 2 functions instead of having these filler properties.

src/host_beta/messageRouter.ts
mliao95: `url` is the parameter name within `args`. I had to rename it to `cssUrl` because the compiler doesn't like that we're redeclaring the `url` const from the `openInEditor` case.

src/devtoolsPanel.ts
mliao95: I think it should be fine as is since none of the following code really depends on this context being set. The only thing this affects is whether the command to enable/disable css mirroring is available to the user which I don't think awaiting or not awaiting will make much of a difference.

src/devtoolsPanel.ts
mliao95: I think we should be fine here.  Tried breaking it a bunch of different ways, but in the end a blank file still starts at line 1.

test/utils.test.ts
vidorteg: thanks for the detailed explanation @mliao95 👌

vidorteg (approved):
thanks for addressing the comments

PR #463, opened 2021-08-18
Create default landing page for webview in case remote devtools fail to load

bgoddar:
This PR adds a default message to the DevTools webview that is only seen if the iframe fails to load anything. This is to address the rare case that the CDN fails on first attempt to debug a new browser version

image

### DIFF
--- a/src/common/styles.css
+++ b/src/common/styles.css
@@ -9,3 +9,15 @@ html, body, iframe {
     margin: 0;
     overflow: hidden;
 }
+
+#error-message {
+    height: 100%;
+    width: 100%;
+    position: absolute;
+    margin: 10px;
+    z-index: 2;
+}
+
+#error-message.hidden {
+    display: none;
+}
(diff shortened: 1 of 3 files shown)


mliao95 (approved):
In general LGTM. Wouldn't mind some better styling for the message even if it is as basic as following a header of "Unable to download DevTools for current target" with potential solutions / actions below.

Example (don't necessarily need the buttons/icons):
image

bgoddar:
> In general LGTM. Wouldn't mind some better styling for the message even if it is as basic as following a header of "Unable to download DevTools for current target" with potential solutions / actions below.
> 
> Example (don't necessarily need the buttons/icons):
> image

Updated image in description.
I've also set a delay of 10 seconds before displaying the error message to prevent confusion on slow devtools loads

src/host_beta/mainHost.ts
```diff
@@ -3,3 +3,7 @@
 
 import { MessageRouter } from './messageRouter';
 export const messageRouter = new MessageRouter(window);
+
+export function iframeFailed() {
```
flynnolivia, line 7: I'm not sure this function is necessary, I don't see it called anywhere?

flynnolivia (approved):
LGTM with a question/nit

src/host_beta/mainHost.ts
bgoddar: Oops! Forgot to remove. Good catch

mliao95:
LGTM after changes.  Thanks!

PR #466, opened 2021-08-19
Update min supported version for CDN tools

bgoddar:
This PR updates the min supported version that the extension will attempt to use the CDN to 94.0.988.0, which is the first canary that has all of the migrated polyfills


flynnolivia (approved):
LGTM

Issue #436, opened 2021-07-19
Clicking a CSS path from inspector fails with remote environment


aaronpowell:
If you open a repo using a VS Code remote environment (issue has been tested on WSL + Containers, assuming it's the same in SSH) and try to click on a link to a CSS file in the inspector (source map generated or otherwise) the file fails to load.

After some debugging I found that the following error was being generated:

```
rejected promise not handled within 1 second: Error: cannot open file:///vscode-remote%3A/wsl%252Bubuntu/home/aaron/code/github/supercharging-web-dev-toolbox/src/pages/CreateGame.module.css. Detail: Unable to read file '\vscode-remote:\wsl%2Bubuntu\home\aaron\code\github\supercharging-web-dev-toolbox\src\pages\CreateGame.module.css' (Error: Unable to resolve non-existing file '\vscode-remote:\wsl%2Bubuntu\home\aaron\code\github\supercharging-web-dev-toolbox\src\pages\CreateGame.module.css')
stack trace: Error: cannot open file:///vscode-remote%3A/wsl%252Bubuntu/home/aaron/code/github/supercharging-web-dev-toolbox/src/pages/CreateGame.module.css. Detail: Unable to read file '\vscode-remote:\wsl%2Bubuntu\home\aaron\code\github\supercharging-web-dev-toolbox\src\pages\CreateGame.module.css' (Error: Unable to resolve non-existing file '\vscode-remote:\wsl%2Bubuntu\home\aaron\code\github\supercharging-web-dev-toolbox\src\pages\CreateGame.module.css')
    at vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code%20Insiders/resources/app/out/vs/workbench/workbench.desktop.main.js:2390:64553
```

My debugging shows that the line that the error comes from is https://github.com/microsoft/vscode-edge-devtools/blob/main/src/devtoolsPanel.ts#L309

_I'll use comments to detail my findings around a possible fix_

aaronpowell:
My debugging has led me to find a few things that, together, might be causing the issue here.

Fundamentally, the `Uri` generated on https://github.com/microsoft/vscode-edge-devtools/blob/main/src/devtoolsPanel.ts#L298 thinks that it's working with a `file` scheme, whereas the scheme really needs to be `vscode-remote` so that it opens using a remote handler.

So, unwinding a little more, the `url` that we are trying to resolve the file on disk is:

```
[...]

PR #467, opened 2021-08-20
Supporting following CSS links on WSL remotes

aaronpowell:
This addresses #436 by doing a check for whether there is a remote, and if so, being more fluid on the way the files are discovered.

I've tested this on WSL and non-remote (Windows) and it worked as expected,


mliao95 (approved):
Set up WSL + Remote workspaces and verified the fix. Thanks for submitting this PR!

bgoddar (approved):
LGTM! thanks so much for contributing this

PR #468, opened 2021-08-20
Fully remove Console Output

bgoddar:
Now that the deprecation message has shipped with 1.3.0, this PR fully removes the console output in favor of the debug console


bgoddar:
I realize this will throw errors in the legacy code, (even if they don't affect users). I think we can hold off on this until the legacy polyfilled devtools are removed

PR #473, opened 2021-08-24
CSS mirroring UI for early feedback

flynnolivia:
This PR
- Adds UI to surface the new feature (#462), give a brief explanation, and prompts the user to provide feedback. This UI is temporary, as we eventually want to move UI for this feature into the DevTools, but will allow us to collect feedback in the mean time.
- Only shows the UI when the user is targeting Canary

mirror-ui-beta

A few remaining questions around the pre-populating the issue that linked for feedback:
- Should we add a label like "feedback" in case the user ends up deleting the pre-populated title?
- Do we want to pre-populate any questions into the issue body? (e.g. asking their ideal workflow with the feature)

### DIFF
--- a/package.json
+++ b/package.json
@@ -561,6 +561,11 @@
                     "name": "Targets"
                 },
                 {
+                    "id": "vscode-edge-devtools-view.mirror-editing",
+                    "name": "CSS Mirror Editing",
+                    "when": "config.vscode-edge-devtools.browserFlavor == Canary"
+                },
+                {
                     "id": "vscode-edge-devtools-view.help-links",
                     "name": "Helpful links"
                 }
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -560,6 +560,11 @@
                     "id": "vscode-edge-devtools-view.targets",
                     "name": "Targets"
                 },
+                {
+                    "id": "vscode-edge-devtools-view.mirror-editing",
+                    "name": "CSS Mirror Editing",
+                    "when": "config.vscode-edge-devtools.browserFlavor == Canary"
```
mliao95, line 566: We probably shouldn't base this off of the `browserFlavor` since it would require us to go back and update it once it hits Dev/Beta/Stable, etc.  This actually reminded me that I needed a browser compatibility check for the vscode commands I registered, so I created a `isCdnHostedTools` context [here](https://github.com/microsoft/vscode-edge-devtools/pull/462/files#diff-95909d339267cba030f94b894f49de695bc6e910f1262da13d851ee866d5109cR460).  You can probably just rebase and use that here as well.

codepo8: +1 on adding the label as it makes it more obvious
-1 on pre-populated questions, as people tend to delete most template content

This is excellent!

codepo8 (commented):
This looks good and with the label it would make things obvious.

PR #474, opened 2021-08-24
Introducing error reporter.

### LINKED ISSUE #85: The extension should have better error logging
Especially due to the new changes where sourcemaps are used to open links back in the editor. Figuring out the correct mapping values to use is difficult.

We should use a similar mechanism to https://github.com/microsoft/vscode-edge-debug2 so that users can tell what url is being mapped to what file (or failing to map in some cases)

vidorteg:
This PR introduces an error reporting mechanism so that the user is aware of:
- feedback for the user from specific actions. e.g if a file is not mapped correctly, let the user know that it requires it's attention.
https://i.imgur.com/wbkEp88.png

- errors or bugs that are not intended behavior and were the user can take action e.g a file that could not be opened because it does not exist anymore
https://i.imgur.com/2DeqrnN.png

For error/bugs the user is prompted to file a bug in github.

Usage will be introduced in a following PR.
Fixes #85

### DIFF
--- a/src/errorReporter.ts
+++ b/src/errorReporter.ts
@@ -7,35 +7,46 @@ export const enum ErrorCodes {
   Informational
 }
 
-export interface ErrorEventInterface{ 
+export interface ErrorEventInterface {
   title: string,
-  message: string,
+  message: string | unknown,
   errorCode: ErrorCodes
 }
 
 export class ErrorReporter {
-    static async showErrorDialog(
-        error: ErrorEventInterface): Promise<void> {
+  static async showErrorDialog(
+    error: ErrorEventInterface): Promise<void> {
 
-        const template = `<!-- Please delete any private information -->
+    const template = `<!-- Please delete any private information -->
         **Stack :**
         ${error.message}
 
         **Additional context:**
-        <!-- Add any other context or about the problem or screenshots here. -->`
+        <!-- Add any other context or about the problem or screenshots here. -->`;
 
-        // cannot do multiline due to:
-        // https://github.com/Microsoft/vscode/issues/48900
-        const answer = await vscode.window
-        .showErrorMessage(
-            `An error occurred:\n${error.title}\n${error.message}`,
-            ...['File a bug', 'Close dialog']
-        );
+    // cannot do multiline due to:
+    // https://github.com/Microsoft/vscode/issues/48900
+    const answer = await vscode.window
+      .showErrorMessage(
+        `An error occurred:\n${error.title}\n${error.message}`,
+        ...['File a bug', 'Close dialog']
+      );
 
-        if(answer === 'File a bug') {
-          const encodedTitle = encodeURIComponent(error.title);
-          const encodedTemplate = encodeURIComponent(template);
-          vscode.env.openExternal(vscode.Uri.parse(`https://github.com/microsoft/vscode-edge-devtools/issues/new?title=[${error.errorCode}]${encodedTitle}&body=${encodedTemplate}&labels=bug`));
-        }
+    if (answer === 'File a bug') {
+      let base = 'https://github.com/microsoft/vscode-edge-devtools/issues/new?';
+      const params: Map<string, string> = new Map<string, string>();
+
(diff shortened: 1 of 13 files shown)


src/errorReporter.ts
```diff
@@ -0,0 +1,68 @@
...
+  static async showErrorDialog(
+    error: ErrorEventInterface): Promise<void> {
+
+    const template = `<!-- Please delete any private information -->
+        **Version:**
+        ${// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+          vscode.extensions.getExtension('ms-edgedevtools.vscode-edge-devtools')?.packageJSON.version || 'N/A'}
+
+        **OS:**
+        <!-- Add your hosting platform e.g Microsoft Windows, WSL -->
```
bgoddar, line 26: nit: I'd also mention MacOS here

src/errorReporter.ts
```diff
@@ -0,0 +1,68 @@
...
+
+        **Additional context:**
+        <!-- Add any other context or about the problem or screenshots here. -->`;
+
+    // cannot do multiline due to:
+    // https://github.com/Microsoft/vscode/issues/48900
+    const answer = await vscode.window
+      .showErrorMessage(
+        `An error occurred:\n${error.title}\n${error.message}`,
+        ...['File a bug', 'Close dialog']
```
bgoddar, line 39: Nit: I think Olivia brought this up as well, wdyt about removing the Close dialog, since there is already a big x button

bgoddar (commented):
I have a couple of nits and a proposal:

I'm wondering if we should add telemetry reporting as part of the ErrorReporter class. A couple of places we report errors after launching an error code. It might make sense to do that as part of the ErrorReporter? The only think is then it would need to be aware of the telemetryReporter instance, so that might prove to be too complicated to make them combine, since this is currently a static class. I'm fine keeping them separate for now

bgoddar:
> I have a couple of nits and a proposal:
> 
> I'm wondering if we should add telemetry reporting as part of the ErrorReporter class. A couple of places we report errors after launching an error code. It might make sense to do that as part of the ErrorReporter? The only think is then it would need to be aware of the telemetryReporter instance, so that might prove to be too complicated to make them combine, since this is currently a static class. I'm fine keeping them separate for now

I actually am leaning towards keeping telemetry out of this, so I'm going to update to approved with some nits

bgoddar (approved):
(updating review to approve with nits commented)

src/errorReporter.ts
```diff
@@ -0,0 +1,68 @@
...
+        ${error.message}
+
+        **Additional context:**
+        <!-- Add any other context or about the problem or screenshots here. -->`;
+
+    // cannot do multiline due to:
+    // https://github.com/Microsoft/vscode/issues/48900
+    const answer = await vscode.window
+      .showErrorMessage(
+        `An error occurred:\n${error.title}\n${error.message}`,
```
bgoddar, line 38: nit: should we even include the '\n' since it doesn't do anything?

vidorteg: Thanks for the review @bgoddar , regarding the telemetry I made the same question myself at some point I even thought one was a superset of the other but eventually got to a similar conclusion; there a couple of high priority scenarios where is better to have them split, so most likely they just have an intersection, so it is better to keep them separate (at least ATM)

src/errorReporter.ts
```diff
@@ -0,0 +1,68 @@
...
+        ${error.message}
+
+        **Additional context:**
+        <!-- Add any other context or about the problem or screenshots here. -->`;
+
+    // cannot do multiline due to:
+    // https://github.com/Microsoft/vscode/issues/48900
+    const answer = await vscode.window
+      .showErrorMessage(
+        `An error occurred: ${error.title}\n${error.message}`,
```
bgoddar, line 38: nit: remove this \n as well

flynnolivia (approved):
This LGTM, +1 to the last nit Brandon mentioned

PR #475, opened 2021-08-24
Error handling implementation

### LINKED ISSUE #436: Clicking a CSS path from inspector fails with remote environment

vidorteg:
[depends on this pull request, splitting to make review easier to understand]
https://github.com/microsoft/vscode-edge-devtools/pull/474

This PR introduces the behavior for handling a common set of errors by using the Error reporter. 

It addresses the case where the mapping files could be mapped but using WSL, the following matrix represents the use cases and expected results

|  Local Mapping exists | Target | Expected result |
|--------------------------|-------|-------------------|
| yes | file | file opens in vscode editor |
| yes | host | file opens in vscode editor |
| no | file | Information dialog: bad mapping |
| no | host | Information dialog: bad mapping |

For any exception on the process mentioned above an `ErrorReporter` error dialog will be created.
Fixes #436

### DIFF
--- a/src/cdpTargetsProvider.ts
+++ b/src/cdpTargetsProvider.ts
@@ -10,6 +10,7 @@ import { fixRemoteWebSocket, getListOfTargets, getRemoteEndpointSettings, IRemot
 import { IncomingMessage } from 'http';
 import https = require('https');
 import { LaunchConfigManager } from './launchConfigManager';
+import { ErrorCodes, ErrorReporter } from './errorReporter';
 
 export class CDPTargetsProvider implements vscode.TreeDataProvider<CDPTarget> {
     readonly onDidChangeTreeData: vscode.Event<CDPTarget | null>;
@@ -139,6 +140,11 @@ export class CDPTargetsProvider implements vscode.TreeDataProvider<CDPTarget> {
                             actualTarget.faviconUrl = filePath;
                             resolve(actualTarget);
                         } catch (e) {
+                            void ErrorReporter.showErrorDialog({
+                                errorCode: ErrorCodes.Error,
+                                title: 'Error obtaining the target favicon',
+                                message: e,
+                            });
                             reject(actualTarget);
                         }
                     });
--- a/src/devtoolsPanel.ts
+++ b/src/devtoolsPanel.ts
@@ -42,6 +42,7 @@ export class DevToolsPanel {
     private consoleOutput: vscode.OutputChannel;
     private timeStart: number | null;
     private devtoolsBaseUri: string | null;
+    private isHeadless: boolean;
 
     private constructor(
         panel: vscode.WebviewPanel,
@@ -57,6 +58,7 @@ export class DevToolsPanel {
         this.config = config;
         this.timeStart = null;
         this.devtoolsBaseUri = this.config.devtoolsBaseUri || null;
+        this.isHeadless = false;
         this.consoleOutput = vscode.window.createOutputChannel('DevTools Console');
         if (config.isJsDebugProxiedCDPConnection) {
             // Direct users to the Debug Console
@@ -99,7 +101,7 @@ export class DevToolsPanel {
         // The browser version is used to select between CDN and bundled tools
         // Future versions of the extension will remove this socket and only use CDN
         this.versionDetectionSocket = new BrowserVersionDetectionSocket(this.targetUrl);
-        this.versionDetectionSocket.on('setBrowserRevision', revision => this.setBrowserRevision(revision));
+        this.versionDetectionSocket.on('setCdnParameters', msg => this.setCdnParameters(msg));
 
 
         // Handle closing
@@ -423,20 +425,21 @@ export class DevToolsPanel {
                 ">
             </head>
             <body>
-                <iframe id="devtools-frame" frameBorder="0" src="${cdnBaseUri}?experiments=true&theme=${theme}"></iframe>
+                <iframe id="devtools-frame" frameBorder="0" src="${cdnBaseUri}?experiments=true&theme=${theme}&headless=${this.isHeadless}"></iframe>
             </body>
             </html>
             `;
     }
 
-    private setBrowserRevision(revision: string) {
-        if (revision !== '') {
+    private setCdnParameters(msg: {revision: string, isHeadless: boolean}) {
+        if (msg.revision !== '') {
             this.config.isCdnHostedTools = true;
-            this.devtoolsBaseUri = `https://devtools.azureedge.net/serve_file/${revision}/vscode_app.html`;
+            this.devtoolsBaseUri = `https://devtools.azureedge.net/serve_file/${msg.revision}/vscode_app.html`;
         } else {
             this.config.isCdnHostedTools = false;
             this.devtoolsBaseUri = '';
         }
+        this.isHeadless = msg.isHeadless;
         this.update();
     }
 
--- a/src/devtoolsPanel.ts
+++ b/src/devtoolsPanel.ts
@@ -60,18 +60,14 @@ export class DevToolsPanel {
         this.devtoolsBaseUri = this.config.devtoolsBaseUri || null;
         this.isHeadless = false;
         this.consoleOutput = vscode.window.createOutputChannel('DevTools Console');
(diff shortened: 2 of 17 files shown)


src/errorReporter.ts
```diff
@@ -0,0 +1,68 @@
+
```
vidorteg, line 1: This class can be skipped as it is being reviewed in a linked PR.

src/cdpTargetsProvider.ts
```diff
@@ -139,6 +140,11 @@ export class CDPTargetsProvider implements vscode.TreeDataProvider<CDPTarget> {
                             actualTarget.faviconUrl = filePath;
                             resolve(actualTarget);
                         } catch (e) {
+                            void ErrorReporter.showErrorDialog({
```
bgoddar, line 143: I don't think this is worth showing an error to the user for. It does not significantly affect workflow, and if it fails on a page with several targets, could get really irritating

src/extension.ts
```diff
@@ -246,10 +255,18 @@ export async function attach(
...
-        } else {
-            telemetryReporter.sendTelemetryEvent('command/attach/error/no_json_array', telemetryProps);
         }
     } while (useRetry && Date.now() - startTime < timeout);
+
+    // If there is no after the timeout then throw an exception
+    if (!responseArray) {
+        void ErrorReporter.showErrorDialog({
+            errorCode: ErrorCodes.Error,
+            title: 'Error while attaching to target',
```
bgoddar, line 265: This seems a bit misleading. Won't this hit if we cant get the available targets? "Error while fetching list of available targets"

src/extension.ts
```diff
@@ -246,10 +255,18 @@ export async function attach(
                     DevToolsPanel.createOrShow(context, telemetryReporter, selection.detail, runtimeConfig);
                 }
             }
-        } else {
-            telemetryReporter.sendTelemetryEvent('command/attach/error/no_json_array', telemetryProps);
         }
     } while (useRetry && Date.now() - startTime < timeout);
+
+    // If there is no after the timeout then throw an exception
```
bgoddar, line 261: nit: 'if there is no response array after the timeout...'

src/extension.ts
```diff
@@ -246,10 +255,18 @@ export async function attach(
...
+    if (!responseArray) {
```
bgoddar, line 262: this probably needs to now be if responseArray.length === 0, since you updated the getListOfTargets to return an empty array instead of undefined on failure? unless [] is falsy. I can't ever remember

src/utils.ts
```diff
@@ -208,23 +209,38 @@ export async function getListOfTargets(hostname: string, port: number, useHttps:
...
-            // Do nothing
+        } catch (e) {
+            // localhost might not be ready as the user might not have a server running
+            // so just show an error dialog for any other condition.
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+            if (e && e.code && e.code === 'ECONNREFUSED'){
+                continue;
+            }
+
+            void ErrorReporter.showErrorDialog({
```
bgoddar, line 227: why do these need to be marked 'void'?

src/devtoolsPanel.ts
```diff
@@ -311,35 +312,42 @@ export class DevToolsPanel {
         // Convert the local url to a workspace path
         const transformer = new debugCore.UrlPathTransformer();
         void transformer.launch({ pathMapping: this.config.pathMapping });
-        const localSource = { path: sourcePath };
-        await transformer.fixSource(localSource);
 
-        sourcePath = localSource.path || sourcePath;
+        // origin in this case is trivial since we expect fixSource to take it out
+        const localSource = { path: sourcePath, origin: 'vscode-webview://' };
```
bgoddar, line 317: nit: Since we expect this to be cleared, I suggest using a totally invalid/nonsense origin just in case, something like 'invalid-origin://' unless that breaks the transformer

src/devtoolsPanel.ts
```diff
@@ -311,35 +312,42 @@ export class DevToolsPanel {
...
+                errorCode: ErrorCodes.Error,
+                title: 'Error while opening file in editor.',
+                message: `${sourcePath} file is not mapped to a local file.`,
+            });
         }
+    }
 
-        // Finally open the document if it exists
-        if (uri) {
+    private async openInEditor(uri:vscode.Uri, line: number, column: number){
```
bgoddar, line 335: nit: I'm not sure why this was broken out into another function, unless you use it somewhere else now?

bgoddar (commented):
Overall Looks good. Thanks for taking this on. I have not had a chance to test locally, but I figured I'd leave the comments I have. Mostly nits and questions for my understanding

src/devtoolsPanel.ts
```diff
@@ -311,35 +312,42 @@ export class DevToolsPanel {
...
+        // per documentation if the file was correctly resolved origin will be cleared.
+        // https://github.com/Microsoft/vscode-chrome-debug-core/blob/main/src/transformers/urlPathTransformer.ts
+        if (!localSource.origin) {
+            // Convert the workspace path into a VS Code url
+            const uri = vscode.Uri.file(localSource.path);
+            await this.openInEditor(uri, line, column);
+        } else {
+            await ErrorReporter.showInformationDialog({
+                errorCode: ErrorCodes.Error,
+                title: 'Error while opening file in editor.',
```
bgoddar, line 329: I actually wouldn't put "Error" in the title here, makes me think something is wrong when in fact it is working like expected. i'd maybe say something like
'Unable to open file in editor'
'${sourcePath} does not map to a local file.'

src/utils.ts
vidorteg: As no response is expected and because VS will end up handling them I think is better to have a fire and forget in here.

src/extension.ts
vidorteg: good catch!

src/devtoolsPanel.ts
vidorteg: I was using it in different places in a previous iteration but I left it as it helps the testing.

src/devtoolsPanel.ts
vidorteg: thanks for all the feedback @bgoddar

flynnolivia (approved):
This LGTM. Thanks for making the ErrorReporter class, this will be easy to reuse in the future!

bgoddar (approved):
Thanks for updating things! LGTM (assuming this will be rebased off your other change first)

PR #481, opened 2021-09-01
Adds instruction on how to include extension version

hxlnt:
Changes iOS as a suggested environment (since I don't believe that one is supported), updates version number sample in #.#.# format, and adds a commented instruction to the user on how to find extension version in VS Code. 

Suggesting this edit because I initially had a hard time finding the version number for the extension when filing a bug, and I accidentally included the version number for the debugger extension instead. 😅


codepo8 (approved):
Looks good to me.

vidorteg (approved):
LGTM, thanks Rachel

PR #497, opened 2021-09-17
Integrate static feedback from webhint

antross:
Compliments the runtime webhint analysis already built into the Issues panel of the DevTools to extend it to source files when running in VSCode.


antross:
@vidorteg I'm keeping this as a draft while I work through a few remaining tasks, but please take a look if you get the chance.

antross:
Closing in favor of #566

PR #513, opened 2021-09-23
Introducing standalone screencast

vidorteg:
Introduces a new screencast experience that:

- Allowed the user to mirror the screencast contents in a standalone tab.
- Enabled mouse events for the screencast view
- Enable emulation for a mobile device.
- Added a setting to toggle on and off standalone screencast: Users can now enable the new experience or fallback to the previous behavior.
- Moved the screencast css and html to a template file
- Added a device list that mirrors the one in devtools, this device list will change the mouse to touch for mobile and touch devices and will stay as a mouse with click events for Desktop and Laptops.
- Integrated the screencast into the existing devtools: the new screencast version can be launched by using the screencast button on the devtools.
- Wire keyboard events so that users can type and use the keyboard in the screencast view.
- Added the option to launch screencast without launching devtools: before a devtools instance was required to launch the new screencast panel, but now a new button (on the target list) allows the user to launch (preview) the url using screencast.


src/screencastPanel.ts
```diff
@@ -0,0 +1,100 @@
...
+    private postToWebview(e: WebSocketEvent, message?: string) {
+        encodeMessageForChannel(msg => this.panel.webview.postMessage(msg) as unknown as void, 'websocket', { event: e, message });
+    }
+
+    private getHtmlForWebview() {
+        const inspectorPath = vscode.Uri.file(path.join(this.extensionPath, 'out/screencast', 'screencast.bundle.js'));
+        const inspectorUri = this.panel.webview.asWebviewUri(inspectorPath);
+		const codiconsUri = this.panel.webview.asWebviewUri(vscode.Uri.joinPath(this.context.extensionUri, 'node_modules', '@vscode/codicons', 'dist', 'codicon.css'));
+        const cssPath = this.panel.webview.asWebviewUri(vscode.Uri.joinPath(this.context.extensionUri, 'out/screencast', 'view.css'));
+        const view = new ScreencastView(this.panel.webview.cspSource, cssPath, codiconsUri, inspectorUri);
```
bgoddar, line 82: nit: since it looks like we create a new view, then call render only once, wdyt about trying to pull the member variables out of ScreencastView and making all of the methods static?

src/utils.ts
```diff
@@ -81,6 +81,7 @@ export const SETTINGS_DEFAULT_HOSTNAME = 'localhost';
 export const SETTINGS_DEFAULT_PORT = 9222;
 export const SETTINGS_DEFAULT_URL = path.resolve(path.join(__dirname, 'startpage', 'index.html'));
 export const SETTINGS_WEBVIEW_NAME = 'Edge DevTools';
+export const SETTINGS_SCREENCAST_WEBVIEW_NAME = 'Edge DevTools';
```
bgoddar, line 84: Not sure the screencast title should be the same as the Devtools? Can we add "preview" or "screencast" to the title?

bgoddar (changes requested):
Left a first round of feedback. Still need to test locally. Additionally, I think it would be good to add some tests for the Screencast folder, but this can be follow up work

src/screencast/screencast.ts
```diff
@@ -0,0 +1,193 @@
...
+        this.width = width;
+        this.height = height;
+        this.updateEmulation();
+    }
+
+    private onUrlKeyDown(event: KeyboardEvent): void {
+        let url = this.urlInput.value;
+        if (event.key === 'Enter' && url) {
+            if(!url.startsWith('http')) {
+                url = 'http://' + url;
```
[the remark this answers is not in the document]
antross: Probably should stick with http since that's what the browser does (the site can upgrade to https, which many do these days).

src/screencast/screencast.ts
```diff
@@ -0,0 +1,193 @@
...
+        const height = this.width;
+        this.width = width;
+        this.height = height;
+        this.updateEmulation();
+    }
+
+    private onUrlKeyDown(event: KeyboardEvent): void {
+        let url = this.urlInput.value;
+        if (event.key === 'Enter' && url) {
+            if(!url.startsWith('http')) {
```
antross, line 180: Probably should also allow the `file:` protocol (though I doubt it's that likely to be directly typed in).

src/screencast/screencast.ts
```diff
@@ -0,0 +1,193 @@
...
+            this.cdpConnection.sendMessageToBackend('Page.navigateToHistoryEntry', {entryId})
+        }
+    }
+
+    private onDomContentEventFired(): void {
+        setTimeout(() => this.updateEmulation(), 100);
+    }
+
+    private onForwardClick(): void {
+        if (this.historyIndex < this.history.length - 1) {
```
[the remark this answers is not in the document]
antross: We already do (though I don't think the CSS styling clearly shows it):
```ts
            this.backButton.disabled = this.historyIndex < 1;
            this.forwardButton.disabled = this.historyIndex >= this.history.length - 1;
```

May need to change opacity/color on the disabled state to make this clear.

src/screencast/screencast.ts
```diff
@@ -0,0 +1,193 @@
...
+        this.updateEmulation();
+        this.updateHistory();
+    }
+
+    private registerInputListeners(): void {
+        for (const eventName of Object.keys(MouseEventMap)) {
+            this.screencastImage.addEventListener(eventName, (event) => {
+                const scale = this.screencastImage.offsetWidth / this.screencastImage.naturalWidth;
+                if (this.isDeviceTouch()) {
+                    this.inputHandler.emitTouchFromMouseEvent(event as MouseEvent, scale);
```
antross, line 88: Currently this keeps "Inspect Element" from the DevTools from highlighting nodes as we hover over them in the preview when emulating a touch device. We should try to send mouse events while the inspect mode is active to allow the overlay to work.

CDP at least tells us when the mode has been cancelled (https://chromedevtools.github.io/devtools-protocol/tot/Overlay/#event-inspectModeCanceled) but I'm not as sure about a CDP event for it being activated (we may have to forward something from the tools).

src/screencast/screencast.ts
```diff
@@ -0,0 +1,193 @@
...
+                this.width = 0;
+                this.height = 0;
+                this.screencastWrapper.classList.add('desktop');
+            } else {
+                const selectedOption = this.deviceSelect[this.deviceSelect.selectedIndex];
+                const deviceWidth = selectedOption.getAttribute('devicewidth');
+                const deviceHeight = selectedOption.getAttribute('deviceheight');
+                if (deviceWidth && deviceHeight) {
+                    this.width = parseInt(deviceWidth);
+                    this.height = parseInt(deviceHeight);
```
antross, line 55: We may want to flip the width/height initially as all the devices currently start in landscape mode, which feels a bit weird since most are phones.

src/utils.ts
vidorteg: I changed it because I got confused by the different names :), usually there is 1:1 association between tabs and extension so having "Edge Devtools Preview" I thought I had a different extension installed (from my own experiments on the hackaton). I don't have a strong opinion for naming so I'm ok either way. I'll bring this to the pod, to hear everyone's thoughts but open to anything :)

src/screencast/screencast.ts
vidorteg: Discussed this offline with @antross, as the new behavior will be off by default (while we stabilize it) we will follow up with this one on another PR to make it easier to review.

package.json
```diff
@@ -690,6 +710,7 @@
         "@types/ws": "7.4.2",
         "@typescript-eslint/eslint-plugin": "4.22.1",
         "@typescript-eslint/parser": "4.22.1",
+        "@vscode/codicons": "^0.0.22",
```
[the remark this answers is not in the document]
vidorteg: done.

src/devtoolsPanel.ts
```diff
@@ -546,4 +557,5 @@ export class DevToolsPanel {
             DevToolsPanel.instance = new DevToolsPanel(panel, context, telemetryReporter, targetUrl, config);
         }
     }
+
```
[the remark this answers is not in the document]
vidorteg: done

src/extension.ts
```diff
@@ -103,6 +109,19 @@ export function activate(context: vscode.ExtensionContext): void {
             const runtimeConfig = getRuntimeConfig();
             DevToolsPanel.createOrShow(context, telemetryReporter, target.websocketUrl, runtimeConfig);
         }));
+
+    context.subscriptions.push(vscode.commands.registerCommand(
+        `${SETTINGS_VIEW_NAME}.toggleScreencast`,
+        (target?: CDPTarget) => {
+            if (!target){
+                telemetryReporter.sendTelemetryEvent('command/screencast/noTarget');
```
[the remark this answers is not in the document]
vidorteg: done

src/screencast/cdp.ts
```diff
@@ -0,0 +1,68 @@
...
+        encodeMessageForChannel(msg => vscode.postMessage(msg, '*'), 'websocket', { message: JSON.stringify(cdpMessage) });
+    }
+
+    registerForEvent(method: string, callback: CdpEventCallback): void {
+        if (this.eventCallbackMap.has(method)) {
+            this.eventCallbackMap.get(method)?.push(callback);
+        }
+        this.eventCallbackMap.set(method, [callback]);
+    }
+}
```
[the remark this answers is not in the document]
vidorteg: done

src/screencast/emulatedDevices.ts
```diff
@@ -0,0 +1,786 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+export const emulatedDevices = [
```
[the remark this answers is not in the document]
vidorteg: done.

src/devtoolsPanel.ts
```diff
@@ -123,6 +124,10 @@ export class DevToolsPanel {
 
         // Handle messages from the webview
         this.panel.webview.onDidReceiveMessage(message => {
+            if ((message as string).includes('toggleScreencast')) {
```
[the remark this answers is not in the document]
vidorteg: done.

src/screencast/screencast.ts
```diff
@@ -0,0 +1,193 @@
...
+        this.deviceSelect.addEventListener('change', () => {
+            if (this.deviceSelect.value.toLowerCase() === 'desktop') {
+                this.width = 0;
+                this.height = 0;
+                this.screencastWrapper.classList.add('desktop');
+            } else {
+                const selectedOption = this.deviceSelect[this.deviceSelect.selectedIndex];
+                const deviceWidth = selectedOption.getAttribute('devicewidth');
+                const deviceHeight = selectedOption.getAttribute('deviceheight');
+                if (deviceWidth && deviceHeight) {
```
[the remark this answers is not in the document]
vidorteg: done.

src/screencast/screencast.ts
```diff
@@ -0,0 +1,193 @@
...
+        });
+
+        this.cdpConnection.registerForEvent('Page.domContentEventFired', () => this.onDomContentEventFired());
+        this.cdpConnection.registerForEvent('Page.frameNavigated', result => this.onFrameNavigated(result));
+        this.cdpConnection.registerForEvent('Page.screencastFrame', result => this.onScreencastFrame(result));
+
+        this.inputHandler = new ScreencastInputHandler(this.cdpConnection);
+
+        this.cdpConnection.sendMessageToBackend('Page.enable', {});
+        let resizeTimeout = 0 as unknown as NodeJS.Timeout;
```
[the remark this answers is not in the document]
vidorteg: done.

src/screencastPanel.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    dispose(): void {
+        ScreencastPanel.instance = undefined;
+
+        this.panel.dispose();
+        this.panelSocket.dispose();
+    }
+
+    private onSocketMessage() {
```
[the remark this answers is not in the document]
vidorteg: done.

src/screencastPanel.ts
```diff
@@ -0,0 +1,100 @@
...
+        // TODO: Handle message
+    }
+
+    private onSocketClose() {
+        this.dispose();
+    }
+
+    private update() {
+        // Check to see which version of devtools we need to launch
```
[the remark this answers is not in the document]
vidorteg: done.

src/screencast/screencast.ts
vidorteg: done.

src/screencast/screencast.ts
vidorteg: done, (I'm now reading "vertical dimensions" instead of "horizontal dimensions" when creating the Element)

src/screencastPanel.ts
vidorteg: I guess we could transform the constructor in a static render method, but I don't see a lot of advantages on it. Keeping it encapsulated in the constructor feels right in this case as the object should not exist without the properties.

src/screencast/screencast.ts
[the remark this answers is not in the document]
vidorteg: Added this as follow up work.

src/screencast/screencast.ts
antross: Interesting. Are any of the horizontal dimensions not an exact flip of the vertical dimensions? If so, we may need a follow-up bug to pick the correct dimensions whether we're horizontal vs. vertical rather than just flipping them like we do today.

src/common/errorCodes.ts
```diff
@@ -0,0 +1,5 @@
+export const enum ErrorCodes {
+  Error,
+  Warning,
+  Informational
+}
```
bgoddar, line 5: nit: newline

antross (approved):
Looks like a good first cut for merging off by default (so much great progress!) 

I like the approach we discussed of reviewing additional fixes in follow-up PRs before turning this on by default since this is getting a bit large.

src/common/webviewEvents.ts
```diff
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 
 export type WebviewEvent = 'getState' | 'getUrl' | 'openInEditor' | 'cssMirrorContent' | 'ready' | 'setState' | 'telemetry' | 'websocket'
-| 'getVscodeSettings' | 'copyText' | 'focusEditor' | 'focusEditorGroup' | 'openUrl' | 'consoleOutput';
+| 'getVscodeSettings' | 'copyText' | 'focusEditor' | 'focusEditorGroup' | 'openUrl' | 'consoleOutput' | 'toggleScreencast' | 'reportError';
```
bgoddar, line 5: instead of adding 'reportError' to the list of webview events, you could do what the DevTools panel does and map reportError to telemetry in the screencast panel:
```
            case 'reportError':
                const [type, message, stack, filename, sourceUrl, lineno, colno] = args;
                this.reportError(type, message, stack, filename, sourceUrl, lineno, colno);
                return true;
```
That way you don't need to update the test to filter out unused events, etc

src/screencast/screencast.ts
```diff
@@ -0,0 +1,201 @@
...
+                this.screencastWrapper.classList.add('desktop');
+            } else {
+                const selectedOption = this.deviceSelect[this.deviceSelect.selectedIndex];
+                const deviceWidth = selectedOption.getAttribute('devicewidth');
+                const deviceHeight = selectedOption.getAttribute('deviceheight');
+                if (deviceWidth && deviceHeight) {
+                    this.width = parseInt(deviceWidth);
+                    this.height = parseInt(deviceHeight);
+                } else {
+                    encodeMessageForChannel(msg => vscode.postMessage(msg, '*'), 'reportError', {
```
bgoddar, line 59: goes with the above comment, but if you just use 'telemetry' here, you don't need to add reportError and then filter it out in the Panel test

src/screencast/screencast.ts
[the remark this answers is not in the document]
bgoddar: I don't see the comment?

src/screencast/view.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    public static getDeviceValueFromTitle(title: string) {
+        return title.replace(/['/' || ' ' || '-']/g, '');
+    }
+
+    public render(): string {
+        const deviceList = this.getDeviceList(emulatedDevices);
+        return this.htmlTemplate(this.webviewCSP, this.cssPath, this.codiconsUri, this.inspectorUri, deviceList);
+    }
+}
```
bgoddar, line 96: newline

src/utils.ts
bgoddar: I'd personally prefer for them to have separate names. If I have both tabs open, I want to be able to see at a glance which is which if neither are focused. I'd be fine with "Edge Devtools" and "Edge Devtools: Screencast" or something like that

bgoddar (approved):
Thanks for addressing the feedback. Left a few more comments, but other than that LGTM

src/screencast/screencast.ts
vidorteg: All of them look like an exact flip.

src/common/errorCodes.ts
vidorteg: done.

PR #524, opened 2021-09-27
Remove deprecated console output

bgoddar:
This PR removes the deprecated console output now that Edge 94 is rolling out to stable

### DIFF
--- a/src/host/toolsHost.ts
+++ b/src/host/toolsHost.ts
@@ -92,10 +92,6 @@ export class ToolsHost {
         encodeMessageForChannel(msg => vscode.postMessage(msg, '*'), 'getVscodeSettings', {id});
     }
 
-    sendToVscodeOutput(consoleMessage: string): void {
-        encodeMessageForChannel(msg => vscode.postMessage(msg, '*'), 'consoleOutput', {consoleMessage});
-    }
-
     copyText(clipboardData: string): void {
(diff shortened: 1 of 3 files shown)


src/host/toolsHost.ts
```diff
@@ -92,10 +92,6 @@ export class ToolsHost {
         encodeMessageForChannel(msg => vscode.postMessage(msg, '*'), 'getVscodeSettings', {id});
     }
 
-    sendToVscodeOutput(consoleMessage: string): void {
```
antross, line 95: Should we remove the polyfills that reference this too then? E.g. in `src/host/polyfills/simpleView.ts` there's a call to `Host.InspectorFrontendHost.sendToVscodeOutput(message);` (also has a related test)

bgoddar: I thought about it, but since the polyfill system should be removed soon, I was trying to avoid touching it

antross: Or perhaps more broadly, is this the time to just purge `src/host` entirely? Or would you prefer to save that for a separate follow-up?

bgoddar: I'd like that to be a separate follow up, so that removing this feature is separate from removing the polyfill system in case a revert is needed

antross (approved):

Issue #501, opened 2021-09-20
Error while fetching list of available targets


lbicknese:
**Version:**
1.3.1

**OS:**
Microsoft Windows

**Stack :**
No available targets to attach.

**Additional Info*

Attempting to launch Angular app using debugger. The project builds and launches Edge window as expected. However, it throws the above error each time.

```json
{
      "name": "ng serve",
      "type": "edge",
      "request": "launch",
      "preLaunchTask": "npm: start",
      "targetTypes": [
        "page"
      ],
      "url": "http://localhost:4200",
      "webRoot": "${workspaceFolder}",
      "sourceMapPathOverrides": {
        "webpack:/*": "${webRoot}/*",
        "/./*": "${webRoot}/*",
        "/src/*": "${webRoot}/*",
        "/*": "*",
        "/./~/*": "${webRoot}/node_modules/*"
      }
    }
```

bgoddar:
Hi, thank you for reporting the issue. From what you described, it sounds like the instance of edge you are launching is not initializing with remote debugging enabled, and the extension is unable to find the active target.

Can you try updating your launch.json config to either:
1.
```
     "name": "ng serve",
      "type": "vscode-edge-devtools.debug", // this will launch edge and attempt to auto attach the DevTools
      "request": "launch",
      "preLaunchTask": "npm: start",
      "targetTypes": [
        "page"
      ],
      "url": "http://localhost:4200",
      "webRoot": "${workspaceFolder}",
      "sourceMapPathOverrides": {
        "webpack:/*": "${webRoot}/*",
        "/./*": "${webRoot}/*",
        "/src/*": "${webRoot}/*",
        "/*": "*",
        "/./~/*": "${webRoot}/node_modules/*"
      }
    }
```
OR
```
     "name": "ng serve",
      "type": "pwa-msedge", // this will launch edge and attempt to auto attach VSCode's JS Debugger, then you can use the inspect icon to attach devtools
      "request": "launch",
      "preLaunchTask": "npm: start",
      "targetTypes": [
        "page"
      ],
      "url": "http://localhost:4200",
      "webRoot": "${workspaceFolder}",
      "sourceMapPathOverrides": {
        "webpack:/*": "${webRoot}/*",
        "/./*": "${webRoot}/*",
        "/src/*": "${webRoot}/*",
        "/*": "*",
        "/./~/*": "${webRoot}/node_modules/*"
      }
    }
```
[...]

PR #560, opened 2021-10-15
Remove legacy Edge Debugger to address potential target list bug

bgoddar:
This PR removes support for the legacy Debugger for Edge extension from the Tools extension. I believe this will help mitigate the occurences of the "No available targets found" error when attempting to use the "attach" entrypoint as described in: https://github.com/microsoft/vscode-edge-devtools/issues/501

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -224,7 +224,8 @@ export async function getListOfTargets(hostname: string, port: number, useHttps:
             if (jsonResponse) {
                 break;
             }
-        } catch (e) {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        } catch (e: any) {
             // localhost might not be ready as the user might not have a server running
             // so just show an error dialog for any other condition.
             // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -228,8 +228,6 @@ export async function getListOfTargets(hostname: string, port: number, useHttps:
             // localhost might not be ready as the user might not have a server running
             // so just show an error dialog for any other condition.
             // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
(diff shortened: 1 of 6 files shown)


bgoddar:
> I'm good with deprecating the legacy Edge Debugger config, but I am still unclear whether this could be the cause of the "fetching list of available targets" bug. Were you able to repro the "fetching list of available targets" error reporter toast without this change and using these config types?

My theory was that our check to see if users had a valid launch.json config was also accepting configs for launching spartan, and then our "launch project" button would attempt to launch spartan edge, and the extension wouldn't have anything to attach to. This is just a theory though. I have not been able to repro.

bgoddar:
@mliao95 or @antross  would one of yall be able to take over this one since I'll be out for awhile?

antross:
@bgoddar after experimenting with various configurations and repros, I decided to leave this support in a bit longer but (1) silenced the request to report an issue for these targets and (2) added a warning when attaching to these targets to migrate to `pwa-msedge` in PR #690.

We'll probably still want to loop back and fully remove these in a future update, though (so leaving your PR active).

src/utils.ts
```diff
@@ -228,8 +228,6 @@ export async function getListOfTargets(hostname: string, port: number, useHttps:
             // localhost might not be ready as the user might not have a server running
             // so just show an error dialog for any other condition.
             // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
Latha303, line 231: // eslint-disable-next-line @typescript-eslint/ban-ts-comment

src/utils.ts
```diff
@@ -228,8 +228,6 @@ export async function getListOfTargets(hostname: string, port: number, useHttps:
...
-            // @ts-ignore specific case for ECONNREFUSED.
```
Latha303, line 232: // @ts-ignore specific case for ECONNREFUSED.

Issue #531, opened 2021-09-29
Dark theme not working


LucasOe:
**Environment (please complete the following information):**
 - OS: Windows 10, Version 21H1
 - Extension version: v1.3.1

**Describe the bug:**
Changing the `vscode-edge-devtools.themes` setting to `dark` doesn't change the theme of the developer tools, even after a restart. Changing the theme using Ctrl+Shit+P also doesn't work.
Note: The addition of new Themes in Microsoft Edge version 93 may has caused this bug (https://docs.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/customize/theme).

**Repro steps:**
1. Change `vscode-edge-devtools.themes` to  `dark`
2. Open a new Microsoft Edge Instance, or attach an existing instance to target

**Expected behavior:**
The dev tools inside VSCode should be displayed in a dark theme.

**Additional context:**
![Code_B0kWjfkyYw](https://user-images.githubusercontent.com/16548335/135293637-c95b3aaf-5e54-42af-a9c1-c86c7c1f3866.png)

Enteleform:
Same issue here.

NeriDarkira:
same issue here, but I think I found a "solution":
in one of the last updates EdgeDevTool started getting the theme directly from the vsCode-theme; so, if you have on of the [available themes](https://docs.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/customize/theme), the EdgeDevTool will "follow" that theme.
if you don't use one of the supported ones, EdgeDevTool will follow the default theme (light).

So if you wanna change the theme you have to change the vsCode theme in one of the supported, 'till they fix this issue ( get default from settings "vscode-edge-devtools.themes" instead of automatically switch to light)

LucasOe:
@NeriDarkira I was able to fix the problem.

[Here](https://github.com/microsoft/vscode-edge-devtools/blob/706bc60c690b9db2932c4595a9ecd5b030827e66/src/devtoolsPanel.ts#L486) the theme gets passed as an URL argument. Passing `Dark` as a capitalized string doesn't work. Using the legacy function `getThemeSettings()` instead of `getThemeFromUserSetting()` and changing `Dark` to `dark` as a setting in the package.json fixes the problem.

This works for now, I'm not sure how to make use of the new dev tools themes introduced in version 93 however.

vidorteg:
I still see this issue on the latest version so reopening it.

[...]

PR #567, opened 2021-10-20
Add DevTool themes and fix theme selection. Refs: #531

LucasOe:
The old getThemeFromUserSetting function would return "systemPreference" instead of using the theme defined in the user settings as a fallback value. I also added all of the available themes as a setting to choose from. The Extension should now work as described in the comment:
1. Fetching the User configured Global VSCode theme, return it if supported
2. Fall back to the extension Theme setting selector
3. Fall back to system preference


src/common/settingsProvider.ts
```diff
@@ -47,7 +48,10 @@ export class SettingsProvider {
   getThemeFromUserSetting(): string {
       const themeSetting = vscode.workspace.getConfiguration().get('workbench.colorTheme');
       const legacySetting = vscode.workspace.getConfiguration(SETTINGS_STORE_NAME).get('themes');
-      return SUPPORTED_THEMES.get((themeSetting || legacySetting) as string) || 'systemPreference';
+      if (SUPPORTED_THEMES.has(themeSetting as string)) {
+        return SUPPORTED_THEMES.get(themeSetting as string) || 'systemPreferred';
+      }
+      return SUPPORTED_THEMES.get(legacySetting as string) || 'systemPreferred';
```
vidorteg, line 54: IMO doing the type casting in the declaration improved readability, what do you think of something like this:
```suggestion
      const themeSetting = vscode.workspace.getConfiguration().get('workbench.colorTheme') as string;
      const legacySetting = (vscode.workspace.getConfiguration(SETTINGS_STORE_NAME).get('themes') || '') as string;
      return SUPPORTED_THEMES.get(themeSetting) || SUPPORTED_THEMES.get(legacySetting) || 'systemPreferred';
```

src/common/settingsProvider.ts
```diff
@@ -5,10 +5,15 @@ import { SETTINGS_STORE_NAME } from '../utils';
 import { ThemeString } from './webviewEvents';
 
 const SUPPORTED_THEMES = new Map<string, string>([
-  ['Default Light+', 'light'],
-  ['Visual Studio Light', 'light'],
+  ['Default Light+', 'default'],
+  ['Visual Studio Light', 'default'],
```
vidorteg, line 9: I don't think this code is required to fix this bug.

src/common/settingsProvider.ts
```diff
@@ -5,10 +5,15 @@ import { SETTINGS_STORE_NAME } from '../utils';
...
   ['Default Dark+', 'dark'],
   ['Visual Studio Dark', 'dark'],
+  ['System Preference', 'systemPreferred'],
```
vidorteg, line 12: same here, not sure if this code is needed.

vidorteg (changes requested):
overall the change looks good just left some comments that need to be addressed before checking it in. Thanks for the contribution 👍

src/common/settingsProvider.ts
LucasOe: The DevTools actually use "default" instead of "light" for the light theme, but since it defaults to the light theme when it doesn't know the string it makes no visible difference. Since I was changing the theme strings anyway, I thought I'd "fix" this, but it's technically not needed.

src/common/settingsProvider.ts
LucasOe: This line is mandatory. The DevTools don't know "systemPreference" as a valid theme.

vidorteg (approved):
Thanks for addressing the feedback. LGTM

codepo8:
@LucasOe Hi there, we would like to merge this PR but you still need to sign the CLA, can you do that? Cheers.

LucasOe:
@codepo8 I have already signed the CLA on the 20th of October 2021, I can't sign it a second time. The first bot message says all CLA requirements met, not sure why the second message says I haven't signed it. Signing it again or rechecking doesn't seem to do anything.

PR #595, opened 2021-10-29
[Screencast] Sync standalone Screencast's device list with in-box Devtools'

hkal:
This PR updates the list of emulated Devices to match the browser DevTools

### DIFF
--- a/src/screencast/emulatedDevices.ts
+++ b/src/screencast/emulatedDevices.ts
@@ -1,4 +1,3 @@
-
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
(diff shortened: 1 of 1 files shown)


src/screencast/emulatedDevices.ts
```diff
@@ -1,10 +1,293 @@
+
```
mliao95, line 1: Do we need this? Also let's add a newline at the end of the file.

mliao95 (approved):
LGTM with some minor comments.

src/screencast/emulatedDevices.ts
hkal: Nope. That was the EOF newline.

PR #604, opened 2021-11-02
[Screencast] Handle user-agent emulation

hkal:
This patch sends `user-agent` overrides to the debug target when device
is changed in standalone screencast.

### DIFF
--- a/src/screencast/screencast.ts
+++ b/src/screencast/screencast.ts
@@ -147,6 +147,12 @@ export class Screencast {
             maxTouchPoints: 1,
         };
 
+        if (this.deviceUserAgent().length > 0) {
+            this.cdpConnection.sendMessageToBackend('Emulation.setUserAgentOverride', {
+                userAgent: this.deviceUserAgent(),
+            });
+        }
+
         this.cdpConnection.sendMessageToBackend('Emulation.setDeviceMetricsOverride', deviceMetricsParams);
         this.cdpConnection.sendMessageToBackend('Emulation.setTouchEmulationEnabled', touchEmulationParams);
         this.toggleTouchMode();
@@ -174,6 +180,14 @@ export class Screencast {
     private isDeviceTouch(){
         const selectedOption = this.deviceSelect[this.deviceSelect.selectedIndex];
         return selectedOption.getAttribute('touch') === 'true' || selectedOption.getAttribute('mobile') === 'true';
+    }
+
+    private deviceUserAgent() {
+        if (this.deviceSelect.value.toLowerCase() === 'desktop') {
+            return navigator.userAgent;
+        }
+        const selectedOption = this.deviceSelect[this.deviceSelect.selectedIndex];
+        return unescape(selectedOption.getAttribute('userAgent') || '');
     }
 
     private updateScreencast(): void {
--- a/src/screencast/screencast.ts
+++ b/src/screencast/screencast.ts
@@ -147,12 +147,9 @@ export class Screencast {
             maxTouchPoints: 1,
         };
 
-        if (this.deviceUserAgent().length > 0) {
-            this.cdpConnection.sendMessageToBackend('Emulation.setUserAgentOverride', {
-                userAgent: this.deviceUserAgent(),
-            });
-        }
-
+        this.cdpConnection.sendMessageToBackend('Emulation.setUserAgentOverride', {
+            userAgent: this.deviceUserAgent(),
+        });
         this.cdpConnection.sendMessageToBackend('Emulation.setDeviceMetricsOverride', deviceMetricsParams);
         this.cdpConnection.sendMessageToBackend('Emulation.setTouchEmulationEnabled', touchEmulationParams);
         this.toggleTouchMode();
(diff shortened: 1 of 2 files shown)


hkal:
nav-test-detached-nodes

hkal:
user-agent-test

src/screencast/view.ts
```diff
@@ -71,7 +71,7 @@ export class ScreencastView {
 
         for (const device of devicesArray) {
             // @ts-ignore ignoring as this is a static template.
-            templatedString += `<option deviceWidth=${device.screen.vertical.width} deviceHeight=${device.screen.vertical.height} ${ScreencastView.getDeviceCapabilities(device.capabilities)} value="${ScreencastView.getDeviceValueFromTitle(device.title)}">${device.title}</option>`;
+            templatedString += `<option deviceWidth=${device.screen.vertical.width} deviceHeight=${device.screen.vertical.height} ${ScreencastView.getDeviceCapabilities(device.capabilities)} userAgent=${escape(device['user-agent'])} value="${ScreencastView.getDeviceValueFromTitle(device.title)}">${device.title}</option>`;
```
hkal, line 74: This is an odd design choice. I would expect us to use a `deviceId` as the select value and lookup the device in `screencast.ts`. Is there a reason we're using custom attributes?

vidorteg: synced with Tony on this one, I don't see any blockers in changing this to used the deviceId. Another possible option as Tony mentioned would be to change the attributes to 'data-' attributes. e.g data-device-width. So I'm onboard with the one that you think will improve readability :-)

vidorteg:
Just noticed something. In the gif posted in this PR you can see that when surface Duo is selected 'Chrome/%s' is shown. AFAIK %s should have the version number. I have not build this PR yet to test it, but just wanted to raise it in case anyone is observing the same.

src/screencast/view.ts
hkal: > So I'm onboard with the one that you think will improve readability
My comment is less about readability and more about ease of extensibility. Currently the view holds the device data model while the viewmodel is the component that will make decisions based on device properties.

src/screencast/screencast.ts
```diff
@@ -176,6 +182,14 @@ export class Screencast {
         return selectedOption.getAttribute('touch') === 'true' || selectedOption.getAttribute('mobile') === 'true';
     }
 
+    private deviceUserAgent() {
+        if (this.deviceSelect.value.toLowerCase() === 'desktop') {
+            return navigator.userAgent;
```
hkal, line 187: I realize this is incorrect. We want the browser's initial userAgent not the VSCode managed webview's.

antross: For that, we should be able to unset the override by passing the empty string.

src/screencast/view.ts
antross: Yes, I also favor putting an ID as the value so we can look up the information from the original source rather than having to copy it through the DOM (and consequently avoid updating this template each time we want to include more data).

src/screencast/screencast.ts
hkal: Yep!

src/screencast/view.ts
antross: @hkal think you can incorporate that into this PR or do you want to break it out as a separate task?

mliao95:
Tested and this LGTM. If we are breaking out the device id change, I'll approve, otherwise, I'll wait on that change.

antross:
@hkal @mliao95 as-is this includes devices that are hidden by default in the in-browser DevTools (e.g. iPhone 5). These appear at the bottom of the list after "Nest Hub Max". Do we want to exclude those to match or leave them included since we don't yet have a way to turn hidden entries on separately?

hkal:
> If we are breaking out the device id change, I'll approve, otherwise, I'll wait on that change.

Refactor will be a different PR/task. I want to keep this scoped to addressing the user-agent bug.

hkal:
> @hkal @mliao95 as-is this includes devices that are hidden by default in the in-browser DevTools (e.g. iPhone 5). These appear at the bottom of the list after "Nest Hub Max". Do we want to exclude those to match or leave them included since we don't yet have a way to turn hidden entries on separately?



> @hkal @mliao95 as-is this includes devices that are hidden by default in the in-browser DevTools (e.g. iPhone 5). These appear at the bottom of the list after "Nest Hub Max". Do we want to exclude those to match or leave them included since we don't yet have a way to turn hidden entries on separately?

Separate bug IMO.

antross (approved):

hkal:
> Just noticed something. In the gif posted in this PR you can see that when surface Duo is selected 'Chrome/%s' is shown. AFAIK %s should have the version number. I have not build this PR yet to test it, but just wanted to raise it in case anyone is observing the same.

The `%s` exists in the [downstream Surface device entries](https://dev.azure.com/microsoft/Edge/_git/chromium.devtools-frontend?path=/front_end/models/emulation/EmulatedDevices.ts&version=GBmaster&line=870&lineEnd=870&lineStartColumn=113&lineEndColumn=123&lineStyle=plain&_a=contents). I'll ask the team.

PR #607, opened 2021-11-04
[Telemetry]: Re-implement screencast telemetry

mliao95:
This PR re-implements these two telemetry events in the new standalone screencast:

  - `DevTools.ScreencastToggle`
      - This fires on each button press and shows whether the view is shown / hidden
          - Show: `devtools/DevTools.ScreencastToggle: {"DevTools.ScreencastToggle.actionCode":"1"}`
          - Hide: `devtools/DevTools.ScreencastToggle: {"DevTools.ScreencastToggle.actionCode":"0"}`
  - `DevTools.ScreencastDuration`
      - This fires when the screencast is toggled off and shows the session time in ms
          - `devtools/DevTools.ScreencastDuration: undefined, {"DevTools.ScreencastDuration.duration":1001}`

### DIFF
--- a/src/screencastPanel.ts
+++ b/src/screencastPanel.ts
@@ -2,30 +2,41 @@
 // Licensed under the MIT License.
 import * as path from 'path';
 import * as vscode from 'vscode';
-import { encodeMessageForChannel, WebSocketEvent } from './common/webviewEvents';
+import {
+    encodeMessageForChannel,
+    WebSocketEvent,
+    ITelemetryProps,
+    ITelemetryMeasures,
+} from './common/webviewEvents';
 import { PanelSocket } from './panelSocket';
 import { ScreencastView } from './screencast/view';
 import {
     SETTINGS_STORE_NAME,
     SETTINGS_SCREENCAST_WEBVIEW_NAME,
 } from './utils';
+import TelemetryReporter from 'vscode-extension-telemetry';
 
 export class ScreencastPanel {
     private readonly context: vscode.ExtensionContext;
     private readonly extensionPath: string;
     private readonly panel: vscode.WebviewPanel;
+    private readonly telemetryReporter: TelemetryReporter;
     private targetUrl: string
     private panelSocket: PanelSocket;
+    private screencastStartTime;
     static instance: ScreencastPanel | undefined;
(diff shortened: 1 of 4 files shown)


src/screencastPanel.ts
```diff
@@ -2,30 +2,41 @@
...
     SETTINGS_STORE_NAME,
     SETTINGS_SCREENCAST_WEBVIEW_NAME,
 } from './utils';
+import TelemetryReporter from 'vscode-extension-telemetry';
 
 export class ScreencastPanel {
     private readonly context: vscode.ExtensionContext;
     private readonly extensionPath: string;
     private readonly panel: vscode.WebviewPanel;
+    private readonly telemetryReporter: TelemetryReporter;
```
hkal, line 23: We should consider making this a singleton that's exported by the telemetry module instead of passing the ref around.

src/screencastPanel.ts
```diff
@@ -47,6 +61,29 @@ export class ScreencastPanel {
...
+    }
+
+    private recordEnumeratedHistogram(actionName: string, actionCode: number) {
+        const properties: ITelemetryProps = {};
+        properties[`${actionName}.actionCode`] = actionCode.toString();
+        this.telemetryReporter.sendTelemetryEvent(
+            `devtools/${actionName}`,
+            properties);
+        // eslint-disable-next-line no-console
+        console.log({name: `devtools/${actionName}`, properties});
```
hkal, line 75: Remove this log and the one in `recordPerformanceHistogram`.

hkal (changes requested):

src/screencastPanel.ts
mliao95: Done

hkal (approved):
Thanks LGTM

src/screencastPanel.ts
mliao95: I like this idea and was going to follow up to ask if you would want to break this out in a separate change, but I guess your approval answers that.

We would probably need to make a wrapper class for the TelemetryReporter since we cannot modify it directly.  We can also move some functions to this wrapper class and cut down on duplicate code.

PR #623, opened 2021-11-09
launch Microsoft Edge with specified args

gp5251:
This PR adds an option to enable cors when launching instance. 

image

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -617,7 +617,7 @@ export function isHeadlessEnabled(): boolean {
 }
 
 /**
- * Verifies if the headless checkbox in extension settings is enabled.
+ * Verifies if the cors checkbox in extension settings is enabled.
  */
 export function isCorsEnabled(): boolean {
     const settings = vscode.workspace.getConfiguration(SETTINGS_STORE_NAME);
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -400,9 +400,14 @@ export async function launchBrowser(browserPath: string, port: number, targetUrl
     ];
 
     const headless: boolean = isHeadlessEnabled();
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -400,11 +400,16 @@ export async function launchBrowser(browserPath: string, port: number, targetUrl
...
 
     const headless: boolean = isHeadlessEnabled();
+    const corsEnabled: boolean = isCorsEnabled();
 
     if (userDataDir) {
         args.unshift(`--user-data-dir=${userDataDir}`);
     }
 
+    if (corsEnabled) {
+        args.unshift('--disable-web-security');
```
hkal, line 410: `--disable-web-security` disables same-origin policy, which isn't the same as CORS iirc. Does launching the browser with this switch address your usecase? Additionally, if CORS is enabled we don't want to pass this switch, so the check should be inversed.

hkal (changes requested):
Thanks for the PR. The team is planning on generalizing passing browser cmdline switches so we don't have to add new extension settings for each one, since there are many. Can you update the PR to work with an extension setting like:
```
"browserArgs" : [
  "--disable-web-security",
  "--user-data-dir=/some/custom/path"
]
```

src/utils.ts
gp5251: yes, `--disable-web-security` works for me.
ok, i changed a bit, using an array as browserArgs now.  should meet the feature.

gp5251: image

hkal: Great. Update the title of the PR to reflect the new work and I'll approve :)

gp5251: updated

hkal (approved):
LGTM

PR #688, opened 2021-11-23
[Screencast]: Register for CDP events when using with js-debug proxy

### LINKED ISSUE #687: Standalone screencast not working when attached via js-debug
 - OS (*e.g.,* Windows 10): Win 10
 - Extension version (*e.g.,* 1.2.3): 1.4.1

**Describe the bug:**
When launching edge via a launch.json debug session, attaching the DevTools works, but subsequently launching the screencast results in a blank display where the screencast content should be.

**Repro steps:**
1. Add a config to launch.json to launch `pwa-msedge`
2. Open DevTools
3. Toggle Screencast

**Expected behavior:**
The page content appears in the screencast

**Additional context:**
Looks like this is because the screencast creates a separate websocket connection that gets proxied through js-debug, but does not request to receive events for various CDP domains. This results in the screencast frame events never making it to the screencast to be displayed.

antross:
Fix #687 

Screencast frames were not being received because the corresponding events were not getting forwarded from js-debug over the proxied CDP connection. Sending the js-debug specific request for events fixes the issue per https://github.com/microsoft/vscode-js-debug/blob/main/CDP_SHARE.md, aligning the screencast panel with the existing behavior for the devtools panel.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -127,7 +127,7 @@ export function activate(context: vscode.ExtensionContext): void {
 
     context.subscriptions.push(vscode.commands.registerCommand(
         `${SETTINGS_VIEW_NAME}.toggleScreencast`,
-        (target?: CDPTarget) => {
+        (target?: CDPTarget, isJsDebugProxiedCDPConnection = false) => {
             if (!target){
                 const errorMessage = 'No target selected';
                 telemetryReporter.sendTelemetryErrorEvent('command/screencast/target', {message: errorMessage});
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -127,15 +127,15 @@ export function activate(context: vscode.ExtensionContext): void {
 
     context.subscriptions.push(vscode.commands.registerCommand(
         `${SETTINGS_VIEW_NAME}.toggleScreencast`,
-        (target?: CDPTarget) => {
+        (target?: CDPTarget, isJsDebugProxiedCDPConnection = false) => {
```
mliao95, line 130: Is there a purpose of defaulting `isJsDebugProxiedCDPConnection` to false here if we already do it in `ScreencastPanel.createOrShow`?

mliao95 (approved):
LGTM with a question.

src/extension.ts
antross: Not particularly. We could just make this an optional param. No strong preference on my part.

PR #690, opened 2021-11-23
Warn when using deprecated 'edge' or 'msedge' launch types

antross:
Also suppress prompting for error reports for missing targets in these cases as switching to `pwa-msedge` is the correct fix.

Displayed warning (shows "edge" or "msedge" depending on actual target type used):
image

The "Learn More" button opens the VS Code documentation on debugging browsers which discusses the correct target types to use: https://code.visualstudio.com/docs/nodejs/browser-debugging

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -387,8 +387,8 @@ export async function attach(
         }
     } while (useRetry && Date.now() - startTime < timeout);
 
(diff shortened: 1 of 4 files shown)


mliao95 (approved):
LGTM.

PR #724, opened 2021-11-29
Silently handle target discovery failures

antross:
A review of the `Error while getting the list of targets` failures shows the vast majority are various forms of connection errors beyond the current check (followed by a few explicit misconfigurations). These are already gracefully handled. Removing this specific report as a result since it is not leading to the discovery of underlying issues in the extension.


mliao95 (approved):
LGTM.  Thanks for the change.

PR #751, opened 2021-12-09
Switch to headless by default

antross:
Updates the default value of the headless setting to launch new Edge instances as headless by default, which will automatically trigger the standalone screencast experience. Also fixes opening the screencast when attaching to an existing instance based on whether the instance is headless (instead of the state of the setting).


ngaitonde (approved):

hkal (approved):
LGTM

PR #803, opened 2022-01-11
fix open folder link to be a button

natashalee:
Open a folder link was solely identifiable using only color, adding a button helps it be more identifiable.


mliao95 (approved):
Tested and LGTM. Thanks!

PR #807, opened 2022-01-11
Refine prompt informing users of errors

antross:
Use a warning as some issues may be transient. Also suggest
checking settings as we've seen some issues are due to
misconfiguration of the extension. Lastly, direct users to a
search of existing issues with the same error message to
discourage filing duplicates.

Example:
image

### DIFF
--- a/src/errorReporter.ts
+++ b/src/errorReporter.ts
@@ -12,43 +12,20 @@ export class ErrorReporter {
   static async showErrorDialog(
     error: ErrorEventInterface): Promise<void> {
 
-    const template = `<!-- Please delete any private information -->
-        **Version:**
-        ${// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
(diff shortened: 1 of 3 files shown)


mliao95:
Generally, LGTM.

One suggestion would be to narrow down the relevant settings for that specific error with more filter terms if possible.  I'm thinking something like this:
image

Also as an FYI, I was having trouble getting `openInEditor` working in Canary.  Getting this error:
image
`Uncaught (in promise) TypeError: Failed to fetch dynamically imported module: https://devtools.azureedge.net/serve_file/@66ea0911636d0b996e4bdf70c832ed0984310f32/entrypoints/vscode_app/VSCodeHost.js`

Could you see if you're hitting this as well?

mliao95 (approved):
LGTM with suggestion above

antross:
I'd like to hold-off on error-specific scoped settings search for now. Not sure the incremental gain is worth the additional work at the moment (curious to see if folks have trouble finding the correct settings once we put this out there).

PR #845, opened 2022-01-31
New stalebot config file

captainbrosset:
This is the config file that the Stale github app can rely on if we decide to use it on this repo: https://github.com/apps/stale

Here is how the app works:

* It will consider an issue as stale after 7 days of inactivity (this way if we haven't responded to someone within a week, we get a ping).
* Issues that are considered stale are given the label `stale`.
* After 7 more days of inactivity, the issue is closed.
* Issues with label `tracked` are never considered stale.


codepo8 (approved):
Looks good to me.

PR #869, opened 2022-02-10
Bug Fix: JSDebug config not properly passed to screencast on initialization

mliao95:
This PR fixes a bug where the screencast does not render anything when first initialized:
vscode

The reason for this is that we are not passing in the `isJsDebugProxiedCDPConnection` parameter when automatically launching the screencast when the browser is in headless mode. The fix for this is the change in `devtoolsPanel.ts`.


hkal (approved):
LGTM - however, there are many changes in this patch that are unrelated to the bug. I suggest they be reverted and staged as a separate patch.

PR #885, opened 2022-02-15
Use webRoot config for source path resolution

### LINKED ISSUE #787: Error while opening CSS file in editor
**Version:**
        1.4.2

        **OS:**
        Windows 10

        **Stack :**
        Could not open document. No workspace mapping was found for 'http://ecalendar-admin-panel/css/styles.css'.

        **Additional context:**
        

## Problem description

I have a project `ecalendar-admin-panel` running on `http://ecalendar-admin-panel:80`. The web root (or "docroot" or whatever) is pointing to `ecalendar-admin-panel/public` folder. Just for clarity, when I go to `http://ecalendar-admin-panel:80` in my browser, I actually request files from `ecalendar-admin-panel/public` folder. 

I use "Launch Edge" configuration to use Edge DevTools. My `launch.json` looks like this:

```json
{
    "configurations": [
        {
            "name": "Launch Edge",
            "request": "launch",
            "type": "pwa-msedge",
            "url": "http://ecalendar-admin-panel:80",
            "webRoot": "${workspaceFolder}/public/",
        }
    ]
}
```

I start a debugging session by pressing this button:

image

And when clicking on `styles.css` in "Edge DevTools" panel I get two errors:

image
image

- ```An error occurred: Error while opening file in editor. Could not open document. No workspace mapping was found for 'http://ecalendar-admin-panel/css/styles.css'..```
- ```Unable to open file in editor. http://localhost/css/styles.css does not map to a local file.```

## Additional info

I've also tried some other `launch.js` configurations, none of which worked, like:

- `"webRoot": "C:/<fullpath>/ecalendar-admin-panel/public/"`
- `"webRoot": "${workspaceFolder}/public"` - No forward slash at the end.
- `"pathMapping": {
    "/": "${workspaceFolder}/public/",
    "http://ecalendar-admin-panel/": "${workspaceFolder}/public/",
  }`

I've also created a folder and a file `css/styles.css` in project root folder (`ecalendar-admin-panel/css/styles.css`) just to test that and then Edge DevTools forwarded me to that file, when I clicked styles.css (i.e. it did work!). But I need Edge DevTools to open the actual `ecalendar-admin-panel/public/css/styles.css` file.

## Suggestion

If I configured everything properly, then "webRoot" and "pathMapping" parameters are ignored by the extension.

bwalderman:
Fixes #787 

The `webRoot` config was not being considered when mapping a URL to a path on the filesystem. This prevented file links in the devtools from successfully opening files in the editor. The fix is to treat `webRoot` as an entry in `pathMapping`.

This fixes the `webRoot` property in Microsoft Edge DevTools' own extension settings, but the extension isn't currently aware of `webRoot` and `pathMapping` properties found in `launch.json` configs. A separate PR https://github.com/microsoft/vscode-js-debug/pull/1206 plumbs those through to our extension so they will work as expected.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -471,6 +471,11 @@ export function getRuntimeConfig(config: Partial<IUserConfig> = {}): IRuntimeCon
         }
     }
 
+    // The webRoot config is equivalent to a pathMapping entry of { '/': '${webRoot}' }.
+    if (!pathMapping.hasOwnProperty('/')) {
+        pathMapping['/'] = webRoot;
+    }
+
     // replace workspaceFolder with local paths
     const resolvedMappingOverrides: IStringDictionary<string> = {};
     for (const customPathMapped in pathMapping) {
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -472,7 +472,7 @@ export function getRuntimeConfig(config: Partial<IUserConfig> = {}): IRuntimeCon
     }
 
     // The webRoot config is equivalent to a pathMapping entry of { '/': '${webRoot}' }.
-    if (!pathMapping.hasOwnProperty('/')) {
+    if (webRoot) {
         pathMapping['/'] = webRoot;
     }
 
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -76,8 +76,8 @@ export function activate(context: vscode.ExtensionContext): void {
         void launch(context);
     }));
 
-    context.subscriptions.push(vscode.commands.registerCommand(`${SETTINGS_STORE_NAME}.attachToCurrentDebugTarget`, (debugSessionId): void => {
-        void attachToCurrentDebugTarget(context, debugSessionId);
+    context.subscriptions.push(vscode.commands.registerCommand(`${SETTINGS_STORE_NAME}.attachToCurrentDebugTarget`, (debugSessionId, config): void => {
+        void attachToCurrentDebugTarget(context, debugSessionId, config);
     }));
 
     // Register the launch provider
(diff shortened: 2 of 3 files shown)


src/utils.ts
```diff
@@ -471,6 +471,11 @@ export function getRuntimeConfig(config: Partial<IUserConfig> = {}): IRuntimeCon
         }
     }
 
+    // The webRoot config is equivalent to a pathMapping entry of { '/': '${webRoot}' }.
+    if (!pathMapping.hasOwnProperty('/')) {
```
antross, line 475: This isn't working for me because I'm getting a pathMapping object back from the `settings.get('pathMapping')` call that already has `{'/': '${workspaceFolder}'}`. However, I don't have that set in my settings.json file so I'm still trying to figure out where it's coming from.

@mliao95 any idea what might be causing this?

mliao95: Not off the top of my head.  I'll dig into it

mliao95: Sorry - got sidetracked, not sure if you already figured it out, but the default value for `pathMapping` is `{'/': '${workspaceFolder}'}` as defined in the `package.json`:
https://github.com/microsoft/vscode-edge-devtools/blob/main/package.json#L217

bwalderman: In that case I think we should always give precedence to `webRoot`. The default `webRoot` is `{$workspaceFolder}`, so this will not affect the default `pathMapping` unless the user has specified something else for `webRoot`.

bwalderman: Thanks @mliao95. Based on the issue report, I think user expectations are that existing launch configs with `webRoot` properties just work with our extension the way they do for vscode JS debugging sessions. On top of that, the launch config should take priority over the extension setting because any logic on the vscode-js-debug side won't even be aware of our extension settings.

mliao95 (approved):

PR #916, opened 2022-03-10
[Feature]: Save last successful DevTools revision to improve reliability

mliao95:
This PR introduces a new setting called `fallbackRevision`.  This setting is updated each time we successfully connect to the DevTools to store the last known working revision. If the DevTools fails to connect to a target, we fallback on using the revision stored in the `fallbackRevision` setting.

This allows users to reliably use the extension offline and when given a faulty revision hash.

### DIFF
--- a/package.json
+++ b/package.json
@@ -318,6 +318,11 @@
                     "type": "boolean",
                     "default": true,
                     "description": "Enable feedback from webhint on source files to improve accessibility, compatibility, security and more."
+                },
+                "vscode-edge-devtools.fallbackRevision": {
+                    "type": "string",
+                    "default": "@c438621746b2932668109129d137a7e3a22ef4fd",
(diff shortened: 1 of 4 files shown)


antross (approved):
LGTM. I tested locally both before/after this change to confirm the improvement. My only thought is that 10s feels a bit long to wait for the fallback, but I'm not sure we can do much better than that with the current hook without risking cutting off successful loads due to a slow network.

hkal (approved):
LGTM - I tested switching channels while offline and it worked as well 👍

mliao95:
> LGTM. I tested locally both before/after this change to confirm the improvement. My only thought is that 10s feels a bit long to wait for the fallback, but I'm not sure we can do much better than that with the current hook without risking cutting off successful loads due to a slow network.

Yeah, I thought the same thing.  At least for the offline scenario it works pretty quickly.

PR #920, opened 2022-03-11
Update  vscode-webhint to version 1.6.5

flynnolivia:
This PR updates the extension to use the latest version of the [vscode-webhint npm package](https://www.npmjs.com/package/vscode-webhint).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,9219 @@
 {
     "name": "vscode-edge-devtools",
     "version": "1.4.5",
-    "lockfileVersion": 1,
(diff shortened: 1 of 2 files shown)


antross:
@mliao95 this PR also bumps the format of package-lock.json due to the update being created with the most recent version of node/npm. I think we should go ahead and cut over, but this will mean future updates to package-lock.json will also need to be created with a newer version of npm to be correct (older versions can still successfully read it, though). See https://github.blog/2021-02-02-npm-7-is-now-generally-available/

PR #932, opened 2022-03-18
[CSS Mirroring] Support CSS Mirroring Enablement from DevTools frontend

hkal:
This patch removes the CSS Mirroring view from the extension panel and
implements the `toggleCSSMirrorContent` host method which toggles the
CSS Mirror content feature in the extensions settings.

### DIFF
--- a/src/devtoolsPanel.ts
+++ b/src/devtoolsPanel.ts
@@ -184,12 +184,7 @@ export class DevToolsPanel {
 
     private onToggleCSSMirrorContent(message: string) {
         const { isEnabled } = JSON.parse(message) as IToggleCSSMirrorContentData;
-
-        if (isEnabled) {
-            void vscode.commands.executeCommand(`${SETTINGS_VIEW_NAME}.cssMirrorContentOn`);
-        } else {
-            void vscode.commands.executeCommand(`${SETTINGS_VIEW_NAME}.cssMirrorContentOff`);
-        }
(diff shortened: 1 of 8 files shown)


src/devtoolsPanel.ts
```diff
@@ -180,6 +182,16 @@ export class DevToolsPanel {
         this.timeStart = performance.now();
     }
 
+    private onToggleCSSMirrorContent(message: string) {
+        const { isEnabled } = JSON.parse(message) as IToggleCSSMirrorContentData;
+
+        if (isEnabled) {
+            void vscode.commands.executeCommand(`${SETTINGS_VIEW_NAME}.cssMirrorContentOn`);
```
mliao95, line 189: We can probably just create one command to toggle it on/off.  We can pass in parameters like this:
```
void vscode.commands.executeCommand(`${SETTINGS_VIEW_NAME}.toggleInspect`, { enabled: false });
```

hkal: Please take another look.

antross (approved):
Tested with Canary and it works great!

Only minor hiccup I found was that state changes sync from the in-devtools checkbox back to the extension setting, but not the other way around if devtools is already open. This is pretty minor as I expect most state changes to occur via the in-devtools checkbox (and devtools correctly reads the current state on launch) don't see it as blocking, but we can log a bug to follow up if we want.

mliao95 (approved):
LGTM

PR #945, opened 2022-03-28
[LaunchConfigManager]: Improve logic for configuring existing launch.json

mliao95:
Issue:
- Due to the LaunchConfigManager changes in https://github.com/microsoft/vscode-edge-devtools/pull/927, configuring existing `launch.json` files would result in either duplicate or missing entries.

Changes:
- Improve logic for checking for missing configs
    - We now cross reference the user's `launch.json` configuration with our expected configurations.  This allows us to add only the missing configs.
    - This improvement is also passed into the logic to determine if a user's `launch.json` configuration is supported or not unsupported.

Things to test:
- Try running `Configure launch.json` against various `launch.json` files.
    - this includes missing `launch.json`, `launch.json` with some of the configurations, `launch.json` with none of the configurations, etc.
image
- With the different `launch.json` files, see which ones generate the `Configure launch.json` button. Assuming that all of the extension configs exist, this button shouldn't show up.

### DIFF
--- a/src/launchConfigManager.ts
+++ b/src/launchConfigManager.ts
@@ -7,7 +7,7 @@ import {
     SETTINGS_STORE_NAME,
     SETTINGS_DEFAULT_URL,
 } from './utils';
-export type LaunchConfig = 'None' | 'Unsupported' | vscode.DebugConfiguration;
+export type LaunchConfig = 'None' | 'Unsupported' | string;
 export type CompoundConfig = {
     name: string,
     configurations: string[],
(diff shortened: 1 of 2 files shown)


src/launchConfigManager.ts
```diff
@@ -165,4 +178,41 @@ export class LaunchConfigManager {
     isValidLaunchConfig(): boolean {
         return this.isValidConfig;
     }
+
+    getMissingConfigs(userConfigs: Record<string, unknown>[], extensionConfigs: Record<string, unknown>[]): Record<string, unknown>[] {
+        const missingConfigs: Record<string, unknown>[] = [];
+        for (const extensionConfig of extensionConfigs) {
+            let configExists = false;
+            for (const userConfig of userConfigs) {
+                if (this.compareConfigs(userConfig, extensionConfig)) {
```
antross, line 187: Because this is effectively a deep comparison, this can end up adding new config entries that mostly duplicate an existing entry if a user has altered a contained value we expect to match (e.g. `--remote-debugging-port`). I see this does allow for additional entries in arrays for things like `runtimeArgs` and allows certain values to be different, though.

The problem occurs when we do hit a non-allowed/incompatible modification, but the name stays the same. In this case end up two entries with the same name after our fixup. Now the compound isn't necessarily going to use the correct one (it appears to pick whichever comes first which is often the incorrect entry). 

I think we want to ensure we also remove/rename the invalid entry in this case.

antross (changes requested):

antross (approved):

PR #963, opened 2022-04-04
Warn and pause mirroring CSS if it would override local edits

antross:
Currently if a user makes local edits to a CSS file, then makes a change in DevTools without saving and rebuilding, the unsaved changes will be overwritten when mirroring CSS content.

This change adds a check to verify the content in the editor either (1) doesn't have any unsaved changes or (2) has only unsaved changes that match what was previously mirrored from DevTools. If these conditions are not met, then the change from DevTools will not be mirrored and a warning notification will be displayed to the user asking them to save and reload the target page (undoing local changes and retrying will also work).

image


mliao95 (approved):
LGTM

PR #966, opened 2022-04-05
Stop debugging session on a headless target when both DevToolsPanel and Screencast are closed

mliao95:
This PR introduces the behavior where the debugging session ends for a headless target when both the DevToolsPanel and Screencast are disposed. This only stops the debugging session if the active session name matches our launch Edge headless config.

Changes:
- add logic to both DevToolsPanel and ScreencastPanel `dispose` methods that will stop debugging session if the target is headless and both panels are closed.

### DIFF
--- a/src/devtoolsPanel.ts
+++ b/src/devtoolsPanel.ts
@@ -150,6 +150,9 @@ export class DevToolsPanel {
                 d.dispose();
             }
         }
+        if (this.isHeadless && !ScreencastPanel.instance) {
+            void vscode.commands.executeCommand('workbench.action.debug.stop');
+        }
     }
 
     private postToDevTools(e: WebSocketEvent, message?: string) {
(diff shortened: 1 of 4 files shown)


antross (commented):
Overall LGTM, but do you think we should apply this to headless non-debug session instances launched from the extension as well (e.g. via the "+" button)?

I'm also still not sure whether it feels right to only scope this to headless or if we should do this regardless of whether the session is headless or not so long as we created it. I'd be curious to hear @codepo8's thoughts.

mliao95:
@antross - Definitely possible to get this working with "+" button, but it isn't that straight forward. It's easier when using a debug session since we can simply just stop the debugging instance.  With the "+" button, we need to either pipe the `browserInstance` or `CdpTargetsProvider` through to the DevToolsPanel and ScreencastPanel.

My thoughts are that + button usage sits at 16% of how users launch a browser/DevTools and it is only going to get smaller as we move in favor of debugging sessions and js-debug. I was planning to leave out this scenario for that reason, but open to adding it if we think it will add enough value moving forward.

antross (approved):

PR #968, opened 2022-04-07
Fix recalculating screencast height when shrinking vertically

antross:
The screencast wrapper previously used to calculate height would be stretched vertically by taller canvas content, such that when shrinking the screencast panel it never correctly shrunk the content.

Fixed by using an element higher up the ancestor chain as the base for sizing. However, this element includes the toolbar so we have to adjust the resulting size by the toolbar height to compensate.


mliao95 (approved):
LGTM.  Thanks for the fix.

PR #975, opened 2022-04-13
[Settings]: Remove theme settings + add auto-detection for light/dark theme for unsupported vscode themes

mliao95:
The extension's DevTools theme would prioritize the user's VSCode theme over the extension's theme settings in VSCode which rendered it useless in most cases.  The only time it would take precedence is if the user had a non-supported vscode theme that we could not apply to the DevTools. To fill in that functionality, this PR falls back to detecting whether the custom theme is light/dark and sets the DevTools theme to the default light/dark theme accordingly.

closes https://github.com/microsoft/vscode-edge-devtools/issues/531

### DIFF
--- a/src/common/settingsProvider.ts
+++ b/src/common/settingsProvider.ts
@@ -2,7 +2,6 @@
 // Licensed under the MIT License.
 import * as vscode from 'vscode';
 import { SETTINGS_STORE_NAME } from '../utils';
-import { ThemeString } from './webviewEvents';
 
 const SUPPORTED_THEMES = new Map<string, string>([
   ['Default Light+', 'default'],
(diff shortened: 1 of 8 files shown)


src/common/settingsProvider.ts
```diff
@@ -34,21 +33,17 @@ export class SettingsProvider {
...
-
   // This function returns the theme for the new frame hosted DevTools by:
   // 1. Fetching the User configured Global VSCode theme, return it if supported
   // 2. Fall back to the extension Theme setting selector (light, dark, system preference)
   // 3. Fall back to system preference
   getThemeFromUserSetting(): string {
       const themeSetting = vscode.workspace.getConfiguration().get('workbench.colorTheme') as string;
-      const legacySetting = (vscode.workspace.getConfiguration(SETTINGS_STORE_NAME).get('themes') || '') as string;
-      return SUPPORTED_THEMES.get(themeSetting) || SUPPORTED_THEMES.get(legacySetting) || 'systemPreferred';
+      let theme = SUPPORTED_THEMES.get(themeSetting);
```
mliao95, line 42: This is pretty much the only additional logic added to this PR.  Everything else is just removing the legacy theme settings.

antross (approved):
Overall LGTM. Worked well with a few additional themes I tried from the marketplace.

I did see one issue we could address here or in a follow-up: when picking any light high-contrast theme (whether built-in or from the marketplace - e.g. GitHub's themes) it always used dark, regardless of whether it was "dark" high-contrast or "light" high-contrast.

src/common/settingsProvider.ts
```diff
@@ -34,21 +33,26 @@ export class SettingsProvider {
...
+      if (!theme) {
+        switch (vscode.window.activeColorTheme.kind as number) {
```
mliao95, line 44: casting this as a number since the ColorThemeKind enum only goes up to 3 even though there are now 4 values.

See https://github.com/microsoft/vscode/blob/c20b68a62d198e73df34de5f95671eb63c8b5ff2/src/vs/workbench/api/common/extHostTypes.ts#L3091

Donald0101:
I need more of this message please.

PR #978, opened 2022-04-14
Fix mapping input coordinates when VS Code is zoomed

### LINKED ISSUE #924: Screencast has cursor offset when VSCode is zoomed in
 - OS (*e.g.,* Windows 10): Windows 11
 - Extension version (*e.g.,* 1.2.3): 1.4.5 (latest)

**Describe the bug:**

When the zoomed in on VSCode, the screencast has an offset from where the cursor is on the screencast and where it is being emulated on the target.

**Repro steps:**

1. Open VSCode
2. Launch an instance of the DevTools using the "+" button in the Edge DevTools extension sidebar menu
3. Zoom in on VSCode by clicking the screencast and pressing CTRL + = multiple times (For Win OS)
4. Attempt to highlight some text in the screencast

**Expected behavior:**
Text should highlight each character as the cursor passes.

**Actual behavior:**
vscode-zoom-bug

**Additional context:**

antross:
Also continues to correctly map coordinates for high DPI displays.

Notably Electron appears to have a long-standing bug where `window.devicePixelRatio` is not always correct. Furthermore, this value (correctly) incorporates both the VS Code zoom level and device zoom level (but the browser only uses the latter). This change avoids issues with this API by deriving the mapping from the difference in sizes between our expected and actual layout width. It also avoids the API when deciding whether to update emulation settings by comparing the ratio of width/height to the expected ratio instead of trying to derive the exact expected dimensions.

Fixes #924

### DIFF
--- a/src/screencast/screencast.ts
+++ b/src/screencast/screencast.ts
@@ -110,7 +110,7 @@ export class Screencast {
 
         for (const eventName of Object.keys(MouseEventMap)) {
             this.screencastImage.addEventListener(eventName, event => {
-                const scale = this.screencastImage.offsetWidth / this.screencastImage.naturalWidth * window.devicePixelRatio;
+                const scale = this.screencastImage.offsetWidth / this.width;
                 const mouseEvent = event as MouseEvent;
                 if (this.isDeviceTouch() && !this.inspectMode) {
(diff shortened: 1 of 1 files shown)


mliao95 (approved):
Verified fix.  LGTM.

PR #981, opened 2022-04-19
Fix scaling down emulated devices to fit available space

antross:
Previously these would end up clipped to the available space, making part of the screen inaccessible.


mliao95 (approved):
Tested and verified.  LGTM

PR #983, opened 2022-04-20
Screencast v2

hkal:
Not ready for review

### DIFF
--- a/src/screencast/flyoutMenuComponent.ts
+++ b/src/screencast/flyoutMenuComponent.ts
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 import {html, render} from 'lit-html';
+import {createRef, ref} from 'lit-html/directives/ref.js'
 import {styleMap, StyleInfo} from 'lit-html/directives/style-map.js';
 
 export enum OffsetDirection {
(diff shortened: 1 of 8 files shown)


src/screencast/flyoutMenuComponent.ts
```diff
@@ -0,0 +1,179 @@
...
+            styles.right = Math.min(document.body.offsetWidth - boundingRect.right).toString() + 'px';
+        }
+
+        render(this.#menuTemplate(styles), document.body);
+        document.body.addEventListener('mousedown', this.#closeMenu)
+
+    };
+
+    #closeMenu = () => {
+        document.getElementById('popover')!.style.display = 'none';
```
antross, line 96: Nit: should we switch popover usages to `ref` as well?

antross (approved):
LGTM (my one nit can be in a follow-up if we decide we want to change to ref). Also here are the other items we discussed offline to tackle in follow-up work:
* Zoom
* Keyboard a11y for flyout menus
* Disable auto-resize if size was manually modified
* Disabled styles for address bar buttons, dimensions inputs, and rotate button

antross:
One other item for follow-up: the current UI overflows when shrinking the screencast to the min-width allowed by VS Code. We should see if we can (a) get things to fit or (b) hide/collapse specific items.

PR #984, opened 2022-04-20
Forward keyboard shortcuts via screencast

antross:
Also fix double-clicking to select text.


mliao95:
It seems to work well when isolated to the context of the screencast, but it doesn't seem to interact with the user's clipboard outside the screencast.

Couple of scenarios that I thought would work but didnt't:
- Copying text inside the screencast and pasting the same text on a text input outside the screencast (in url bar, vscode, etc.)
- Copying text from outside the screencast and pasting that text inside the screencast

antross:
> It seems to work well when isolated to the context of the screencast, but it doesn't seem to interact with the user's clipboard outside the screencast.

Thanks for debugging this with me! 

After @mliao95 and I dug into this a bit we identified that the clipboard for headless instances is isolated from the rest of the system. The behavior works as expected when using a non-headless instance. Given headless is our recommended default I'm going to dig into this some more to see if there's a way we can (1) get headless to use the system clipboard or (2) add logic on our end to copy data between the two clipboard instances.

mliao95:
Tested and it looks good. For the scope of this PR, it LGTM.

There are some browser based shortcuts that we might want to consider porting:
- Find: CTRL + F
- Refresh: CTRL + R, F5
- Back/Forward: ALT + Left/Right

And some lesser important ones:
- Print: CTRL + P
- Focus URL bar: CTRL + L

mliao95 (approved):

PR #985, opened 2022-04-20
Switch to using ref instead of id

antross:
Based on https://lit.dev/docs/templates/directives/#ref


hkal (approved):
Much cleaner way to get the HTMLElement. Thanks for looking it up :)

PR #986, opened 2022-04-20
[Telemetry]: Add telemetry for webhint hover events

mliao95:
This PR adds telemetry events for when a user hovers over a webhint issue.

This is an example event:
`user/webhint/hover: {"hint":"axe/language"}`


src/utils.ts
```diff
@@ -797,6 +797,31 @@ export async function reportFileExtensionTypes(telemetryReporter: Readonly<Telem
...
+        }
+        return true;
+    }
+    return false;
+}
+
+export function getSupportedStaticAnalysisFileTypes(): string[] {
+    const supportedFileTypes = [];
+    for (const event of packageJson.activationEvents) {
+        if (event.includes('onLanguage:')) {
```
antross, line 818: We can tighten this check a bit more, especially since we'll be stripping the prefix if it matches.
```suggestion
        if (event.startsWith('onLanguage:')) {
```

antross (approved):

PR #990, opened 2022-04-25
Fix resizing and scaling screencast

antross:
Cleaned up some CSS so heights would be consistent to avoid pushing the emulation bar down (and some minor additional touches to align the top toolbar more closely with the design mocks). Plumbed offsets for the emulation and toolbar heights through the dimension component when in responsive mode. Set the correct initial responsive state and width/height.

### DIFF
--- a/src/screencast/dimensionComponent.ts
+++ b/src/screencast/dimensionComponent.ts
@@ -15,7 +15,7 @@ export default class DimensionComponent {
     width: number;
     height: number;
     heightOffset: number;
-    updateOnResize = false;
+    updateOnResize = true;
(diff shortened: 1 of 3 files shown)


src/screencast/dimensionComponent.ts
```diff
@@ -32,6 +32,7 @@ export default class DimensionComponent {
         window.addEventListener('resize', this.#onResize.bind(this));
 
         this.update();
+        this.onUpdateDimensions(this.width, this.height);
```
hkal, line 35: Maybe `update()` should just call this function

hkal (approved):
LGTM

PR #991, opened 2022-04-26
Add an inspect button to open DevTools from the screencast

antross:


antross:
@hkal I went ahead and created this targeting your screencast-v2 PR, but I figure we'll get that merged first. Then we can retarget this to main.

hkal (approved):
I gave this patch a go locally and it works. Although, I have a couple comments about the UX we may want to consider:
  - We should disable the open button if DevTools is already open
  - Maybe this should be a toggle like "inspect element"

PR #999, opened 2022-04-28
Servicing 1.4.8

hkal:


antross (approved):
LGTM - we currently have rebase merge disabled, but perhaps @mliao95 can override for this PR? I don't think we want to squash these particular changes.

Also, are you planning to bump the package version and update the changelog in a separate PR?

mliao95:
Enabled rebase merging

hkal:
> LGTM - we currently have rebase merge disabled, but perhaps @mliao95 can override for this PR? I don't think we want to squash these particular changes.
> 
> Also, are you planning to bump the package version and update the changelog in a separate PR?

I included the changelog patch in this PR. I don't think we need an additional one.

PR #1005, opened 2022-05-02
Use Edge 102 DevTools as min version for sourcemap support

antross:
Use the revision of DevTools from the current Edge Dev channel for all targets less than 102.
Ensures CSS mirroring across source-maps is present regardless of the target version of Edge.
Newer versions of Edge will continue to use their own DevTools revision.

### DIFF
--- a/src/versionSocketConnection.ts
+++ b/src/versionSocketConnection.ts
@@ -15,8 +15,9 @@ export interface BrowserVersionCdpResponse {
    }
 }
 
-// Minimum supported version of Edge for new CDN system
-export const MIN_SUPPORTED_VERSION = '94.0.988.0';
+// Minimum supported version of Edge for source-mapped CSS mirroring
+export const MIN_SUPPORTED_VERSION = '102.0.1235.1';
+export const MIN_SUPPORTED_REVISION = '@a41122be052fe9616f0def5fe6842fa942930d46';
 
 export class BrowserVersionDetectionSocket extends EventEmitter {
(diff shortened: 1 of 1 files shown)


src/versionSocketConnection.ts
```diff
@@ -79,7 +80,7 @@ export class BrowserVersionDetectionSocket extends EventEmitter {
                 return {revision: currentRevision, isHeadless};
             }
             if (currentVersion[i] < minSupportedVersion[i]) {
-                return {revision: '', isHeadless};
+                return {revision: MIN_SUPPORTED_REVISION, isHeadless};
```
hkal, line 83: We may want to another check for Edge 101 and display an upgrade prompt/error message if the target is below it, since some extension functionality was moved to dtfe.

hkal (approved):
LGTM - left a non-blocking comment for consideration.

mliao95 (approved):
LGTM

src/versionSocketConnection.ts
antross: Thanks. Anything you can think of in CDP that would be a problem here? I tested this forced upgrade with both Edge 101 and Edge 100 - DevTools worked fine for both since we're now using the 102 dt-fe code in either case. Barring a breaking CDP change I think we should be able to leave this as-is.

hkal: Nope. CDP compat should be fine for a few major browser versions back.

PR #1010, opened 2022-05-03
Use 'known good' revision for all current Edge versions

antross:
Avoids a known issue with CSS source-mapped mirroring in Canary.
A follow-up PR will narrow the range once a new Canary is available with a fix.


mliao95 (approved):
LGTM

PR #1012, opened 2022-05-04
Add telemetry for emulation states

hkal:


src/screencastPanel.ts
```diff
@@ -137,6 +139,19 @@ export class ScreencastPanel {
...
     }
 
+    private onSocketTelemetry(message: string) {
+        const telemetry: TelemetryData = JSON.parse(message) as TelemetryData;
+        if (telemetry.event !== 'screencast') {
+            return;
+        }
+
+        const properties: ITelemetryProps = {};
+        properties[telemetry.name] = JSON.stringify(telemetry.data);
```
antross, line 149: @hkal I see this is based on how we submit errors, but will we be able to query / create summaries based on these stringified pieces of data?

```
devtools/Screencast.Emulation: {"Screencast.Emulation":"{\"event\":\"forcedColors\",\"value\":\"active\"}"}, undefined
```

mliao95: Similar comment.  Would prefer to have the data formatted like this:
```
devtools/Screencast.Emulation: {"event":"forcedColors","value":"active"}, undefined
```

mliao95:
Taking another look at this, I realize that the telemetry event structure is kind of odd.

We shouldn't be categorizing a telemetry event using an "event" property, it should be part of the event itself.

For instance, I would imagine an event like this:

```
devtools/Screencast.Emulation: {"event":"emulatedMedia","value":"screen"}, undefined
```

should be structured like this:
```
devtools/Screencast.Emulation/emulatedMedia: {"value":"screen"}, undefined
```

mliao95 (approved):
LGTM.  Thanks for the changes.

PR #1016, opened 2022-05-05
Remove standalone screencast setting and related code

antross:
Effectively removes the old screencast in favor of always using the standalone screencast implementation.

### DIFF
--- a/src/devtoolsPanel.ts
+++ b/src/devtoolsPanel.ts
@@ -487,12 +487,6 @@ export class DevToolsPanel {
 
         const theme = SettingsProvider.instance.getThemeFromUserSetting();
         const cssMirrorContent = SettingsProvider.instance.getCSSMirrorContentSettings();
-        const standaloneScreencast = SettingsProvider.instance.getScreencastSettings();
-
-        // The headless query param is used to show/hide the DevTools screencast on launch
-        // If the standalone screencast is enabled, we want to hide the DevTools screencast
-        // regardless of the headless setting.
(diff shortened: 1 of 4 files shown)


src/devtoolsPanel.ts
```diff
@@ -514,7 +508,7 @@ export class DevToolsPanel {
                 ">
             </head>
             <body>
-                <iframe id="devtools-frame" frameBorder="0" src="${cdnBaseUri}?experiments=true&theme=${theme}&headless=${enableScreencast}&standaloneScreencast=${standaloneScreencast}&cssMirrorContent=${cssMirrorContent}"></iframe>
+                <iframe id="devtools-frame" frameBorder="0" src="${cdnBaseUri}?experiments=true&theme=${theme}&standaloneScreencast=true&cssMirrorContent=${cssMirrorContent}"></iframe>
```
hkal, line 511: We should stage a patch to deprecate the `standaloneScreencast` queryparam in dtfe

hkal (approved):
LGTM

src/devtoolsPanel.ts
antross: @hkal don't we need to keep that param for the DevTools Visual Studio extension (not VS Code)? This extension continues to opt-in so other hosts that don't have their own standalone screencast can use the one built-in with DevTools.

hkal: Oh right - totally slipped my mind. No followup is needed then.

PR #1017, opened 2022-05-05
Remove obsolete network setting and related code

antross:
This setting was already a no-op. Removing to avoid confusion.


hkal (approved):
LGTM

PR #1020, opened 2022-05-09
Remove obsolete welcome setting and related code

antross:
This setting was already a no-op. Removing to avoid confusion.

### DIFF
--- a/package.json
+++ b/package.json
@@ -249,11 +249,6 @@
                     "description": "Show service and shared workers in the target list.",
                     "default": false
                 },
-                "vscode-edge-devtools.welcome": {
-                    "type": "boolean",
-                    "description": "Enable Welcome panel (requires relaunching extension)",
-                    "default": true
-                },
                 "vscode-edge-devtools.headless": {
                     "type": "boolean",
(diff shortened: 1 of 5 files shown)


test/utils.test.ts
```diff
@@ -1028,7 +1027,7 @@ describe("utils", () => {
         it('correctly records all changed extension settings', async () => {
             const reporter = createFakeTelemetryReporter();
             utils.reportExtensionSettings(reporter);
-            expect(reporter.sendTelemetryEvent).toBeCalledWith('user/settingsChangedAtLaunch', { welcome: 'true', enableNetwork: 'false' });
+            expect(reporter.sendTelemetryEvent).toBeCalledWith('user/settingsChangedAtLaunch', { enableNetwork: 'false' });
```
hkal, line 1030: Since we have a patch that removes `enableNetwork` we should choose something else to use in test code.

hkal (changes requested):

test/utils.test.ts
antross: Agreed. In fact, I switched this to use `isHeadless` in the `enableNetwork` patch already :)
image

PR #1021, opened 2022-05-09
Remove support for legacy 'edge' and 'msedge' debugger types

antross:
The `edge` and `msedge` target types should no longer be used now that the "Debugger For Microsoft Edge" extension has been removed from VS Code in favor of jsdebug (which uses `pwa-msedge`). While the legacy extension was still in the marketplace we warned about the deprecation of these target types. Now that it's been removed we can remove support entirely.

### DIFF
--- a/package.json
+++ b/package.json
@@ -229,11 +229,6 @@
                     "description": "Use JavaScript source maps (if they exist)",
                     "default": true
                 },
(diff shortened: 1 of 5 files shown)


leotlee (approved):

hkal (approved):
LGTM

PR #1022, opened 2022-05-09
Move CSS mirror editing status from settings to globalState

antross:
Leaves the checkbox in the DevTools as the primary UI for toggling state.
Also keeps the command for toggling state and fixes it to work when invoked from the command menu.


hkal (approved):
Code looks fine. However, I still think this state makes sense as a setting.

mliao95:
Given that we see little settings page interaction and we have a more visible checkbox for the user to toggle, I think we can remove it from the settings page.

PR #1024, opened 2022-05-09
Move CDN fallback revision from settings to globalState

antross:
This primarily needs to hold the most recent successful hash and does not need to be configured by the user.


hkal (approved):
LGTM

PR #1028, opened 2022-05-11
webhint quick-fixes support for vscode-edge-devtools

vidorteg:
<TBD>

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -34,6 +34,10 @@ import {
     reportChangedExtensionSetting,
     reportExtensionSettings,
     reportUrlType,
+    getWehbhintConfigPath,
+    ignoreProblemInHints,
+    ignoreCategoryPerProject,
+    ignoreCategoryGlobally,
 } from './utils';
 import { LaunchConfigManager } from './launchConfigManager';
 import { ErrorReporter } from './errorReporter';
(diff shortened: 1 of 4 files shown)


src/extension.ts
```diff
@@ -46,17 +52,20 @@ import {
...
 let telemetryReporter: Readonly<TelemetryReporter>;
 let browserInstance: Browser;
 let cdpTargetsProvider: CDPTargetsProvider;
+let globalStoragePath: string | undefined;
+
+// List of document types the extension will run against, supported documents
+// need a scheme, so adding support for 'file' and 'untitled'(in memory)
+const supportedFiles = getSupportedStaticAnalysisFileTypes().map(item => { return { language: item, scheme: 'file' }; });
+const supportedUntitled = getSupportedStaticAnalysisFileTypes().map(item => { return { language: item, scheme: 'untitled' }; });
+const supportedDocuments = [...supportedFiles, ...supportedUntitled];
```
antross, line 66: The `scheme` is not needed when just using a string instead of an object per https://code.visualstudio.com/api/references/vscode-api#DocumentSelector, so we can just use the return value of `getSupportedStaticAnalysisFileTypes()` directly:

> Note that a document selector that is just a language identifier selects all documents, even those that are not saved on disk.

src/extension.ts
```diff
@@ -293,6 +344,9 @@ function startWebhint(context: vscode.ExtensionContext): void {
             client.outputChannel.clear();
             client.outputChannel.show(true);
         });
+
+        // Once client is ready reload any present configuration.
+        client.sendNotification(reloadAllProjectsConfigNotification);
```
antross, line 349: Since this effectively restarts the language server anyway, could we do the following to avoid having an extra contract between the host and language server?
```ts
await stopWebhint();
startWebhint(context);
```

src/extension.ts
```diff
@@ -271,6 +281,47 @@ function startWebhint(context: vscode.ExtensionContext): void {
...
+                    const hintName = args[1] as string | undefined;
+
+                    switch (command) {
+                        case 'vscode-webhint/ignore-hint-project': {
+                            void ignoreHintPerProject(hintName);
+                            break;
+                        }
+                        case 'vscode-webhint/ignore-hint-global': {
+                            await ignoreHintGlobally(hintName, globalStoragePath || '');
+                            client.sendNotification(reloadAllProjectsConfigNotification);
```
antross, line 296: Since this effectively restarts the language server anyway, could we do the following to avoid having an extra contract between the host and language server (and wrap it as a `restartWebhint` helper)?
```ts
await stopWebhint();
startWebhint(context);
```

src/webhintUtils.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  try {
+      // verify if .hintrc exists
+      await vscode.workspace.fs.stat(path);
+  } catch (error) {
+      if ((error as vscode.FileSystemError).code === vscode.FileSystemError.FileNotFound.name) {
+
+          // .hintrc does not exists so create one with the default config
+          const defaultConfig = { extends: ['development'] };
+          await vscode.workspace.fs.writeFile(path, Buffer.from(JSON.stringify(defaultConfig), 'utf8'));
```
antross, line 25: I wonder if we should do these as workspace edits to maximize VS Code's understanding of what we're doing. See this sample extension for an example: https://github.com/microsoft/vscode-extension-samples/blob/30c1740717acdc7e4aabc783d07bd83e21857320/code-actions-sample/src/extension.ts#L74

PR #1032, opened 2022-05-13
[Screencast]: Change Open DevTools button to Toggle DevTools

### LINKED ISSUE #996: Make Screencast DevTools button "toggle" instead of "open"
I gave this patch a go locally and it works. Although, I have a couple comments about the UX we may want to consider:
  - We should disable the open button if DevTools is already open
  - Maybe this should be a toggle like "inspect element"

_Originally posted by @hkal in https://github.com/microsoft/vscode-edge-devtools/pull/991#pullrequestreview-956769418_

mliao95:
This PR changes the "Open DevTools" button on the screencast to be "Toggle DevTools"

DevTools Closed:
image

DevTools Open:
image

closes #996

### DIFF
--- a/src/screencast/view.ts
+++ b/src/screencast/view.ts
@@ -40,7 +40,7 @@ export class ScreencastView {
                   <i class="codicon codicon-refresh"></i>
               </button>
               <input id="url" />
-              <button id="inspect" title="Toggle DevTools">
+              <button id="inspect" title="${isDevToolsOpen ? 'Close DevTools' : 'Open DevTools'}">
                   <i class="codicon codicon-inspect ${isDevToolsOpen ? 'devtools-open' : ''}"></i>
               </button>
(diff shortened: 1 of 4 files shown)


src/screencast/view.ts
```diff
@@ -39,8 +40,8 @@ export class ScreencastView {
                   <i class="codicon codicon-refresh"></i>
               </button>
               <input id="url" />
-              <button id="inspect" title="Open DevTools">
-                  <i class="codicon codicon-inspect"></i>
+              <button id="inspect" title="Toggle DevTools">
```
antross, line 43: Nit: Should we have this switch between "Open DevTools" and "Close DevTools" based on state (since we have it right here)?

antross (approved):

antross (approved):

PR #1035, opened 2022-05-16
[Screencast]: Add text-only clipboard functionality

### LINKED ISSUE #998: Sync copy/paste from headless screencast with system clipboard
Recent work has enabled many more keyboard shortcuts to work via the embedded browser preview (screencast) provided by the extension. However, when debugging a headless instance, it uses its own isolated clipboard (due to https://bugs.chromium.org/p/chromium/issues/detail?id=697559).

We can make this experience better by synchronizing text between the headless instance and the system clipboard when using the screencast.

mliao95:
This PR implements text-only clipboard functionality to the screencast. We achieve this functionality in this PR through these steps:

Scenario: Copying text from the screencast and pasting it outside screencast
1. We add a `clipboardRequests` set to `cdp.ts` to track request/response ids of CDP messages that relate to clipboard interaction
2. We register a clipboard handler function called `onSaveToClipboard` from `screencast.ts` on `cdp.ts` to handle incoming clipboard responses
3. We send a CDP call to `Runtime.evaluate` and run `document.getSelection().toString()` to request the selected text and store the request ID to `clipboardRequests`
4. When we get a response ID that matches an ID in `clipboardRequests`, we call the `onSaveToClipboard` function and delete the ID from the `clipboardRequests` set.
5. `onSaveToClipboard` posts the message for `ScreencastPanel` to listen and copy to the system keyboard via `vscode.env.clipboard.writeText(s)` 

Scenario: Copying text from outside the screencast and pasting in the screencast
1.  When detecting a paste keyboard shortcut, we make a request to `ScreencastPanel` to send over clipboard contents
(Get shortcut -> call `this.cdpConnection.readClipboardAndPasteRequest()` / `getClipboardContents()`)
2. `ScreencastPanel` reads the system clipboard and sends message back with `readClipboard` event name.
3. `cdpConnection` receives `readClipboard` event and calls registered callback which is `pasteClipboardContents()`
4. `pasteClipboardContents()` sends a `Runtime.evaluate` message to insert the received clipboard text to the current user selection.

closes #998

### DIFF
--- a/src/screencast/input.ts
+++ b/src/screencast/input.ts
@@ -52,6 +52,13 @@ export class ScreencastInputHandler {
             keyboardEvent.preventDefault();
             keyboardEvent.stopPropagation();
         }
+        if (keyboardEvent.ctrlKey && (keyboardEvent.key === 'c' || keyboardEvent.key === 'x') && keyboardEvent.type === 'keydown') {
+            this.cdpConnection.sendMessageToBackend('Runtime.evaluate', {
+                expression: 'document.getSelection().toString()',
+            }, (text: {result: {value: string}}): string => {
+                return text.result.value;
+            }, true);
+        }
         if (keyboardEvent.type === 'keydown' || keyboardEvent.type === 'keyup') {
             const text = hasNonShiftModifier ? '' : this.textFromEvent(keyboardEvent);
             this.cdpConnection.sendMessageToBackend('Input.dispatchKeyEvent', {
@@ -63,7 +70,7 @@ export class ScreencastInputHandler {
                 modifiers: this.modifiersForEvent(keyboardEvent),
                 windowsVirtualKeyCode: keyboardEvent.keyCode,
                 nativeVirtualKeyCode: keyboardEvent.keyCode,
-                text
+                text,
             });
         }
     }
--- a/src/screencast/input.ts
+++ b/src/screencast/input.ts
@@ -55,11 +55,11 @@ export class ScreencastInputHandler {
         if (keyboardEvent.ctrlKey && (keyboardEvent.key === 'c' || keyboardEvent.key === 'x') && keyboardEvent.type === 'keydown') {
             this.cdpConnection.sendMessageToBackend('Runtime.evaluate', {
                 expression: 'document.getSelection().toString()',
-            }, (text: {result: {value: string}}): string => {
-                return text.result.value;
-            }, true);
+            }, undefined, true);
         }
-        if (keyboardEvent.type === 'keydown' || keyboardEvent.type === 'keyup') {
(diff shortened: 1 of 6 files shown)


src/screencast/input.ts
```diff
@@ -52,6 +52,13 @@ export class ScreencastInputHandler {
             keyboardEvent.preventDefault();
             keyboardEvent.stopPropagation();
         }
+        if (keyboardEvent.ctrlKey && (keyboardEvent.key === 'c' || keyboardEvent.key === 'x') && keyboardEvent.type === 'keydown') {
```
antross, line 55: We'll need to handle `metaKey` too for this to work as expected on Macs (not sure if we'll need to make this conditional based on platform).
```suggestion
        if ((keyboardEvent.ctrlKey || keyboardEvent.metaKey) && (keyboardEvent.key === 'c' || keyboardEvent.key === 'x') && keyboardEvent.type === 'keydown') {
```

src/screencast/input.ts
```diff
@@ -52,6 +52,13 @@ export class ScreencastInputHandler {
...
+            this.cdpConnection.sendMessageToBackend('Runtime.evaluate', {
+                expression: 'document.getSelection().toString()',
+            }, (text: {result: {value: string}}): string => {
+                return text.result.value;
```
antross, line 59: Does this `return` do anything? It looks like your changes to `cdp.ts` read the value directly out of the message and don't look at the return value from the callback.

antross (changes requested):
When testing, this worked for copying data from the screencast to the system clipboard, but I still couldn't copy new text from outside the screencast and paste it in. I'd expect to see a call to `vscode.env.clipboard.readText` somewhere to support this second case.

src/screencast/screencast.ts
```diff
@@ -388,6 +399,16 @@ export class Screencast {
         });
     }
 
+    private getClipboardContents(): void {
+        encodeMessageForChannel(msg => vscode.postMessage(msg, '*'), 'readClipboard');
+    }
+
+    private pasteClipboardContents(message: string) {
+        this.cdpConnection.sendMessageToBackend('Runtime.evaluate', {
+            expression: `document.execCommand("insertText", false, "${message}");`,
```
antross, line 408: We should escape any double-quote `"` characters in `message` to avoid a script error in the `Runtime.evaluate` call.

antross (changes requested):
One last bug to address, but after that this should be ready to go.

src/screencast/screencast.ts
mliao95: Good catch.  Fixed in latest patch.

antross (approved):
LGTM - thanks for the updates!

PR #1040, opened 2022-05-19
[Bug fix]: Route no config debug session to default launch experience

mliao95:
Issue:
- If a user does not have an existing `launch.json` and selects `Microsoft Edge Tools` as the debug session, we display a `"No supported launch config was found"` error message.

Changes:
- Handle the no-config case in `launchDebugProvider.ts` similar to this mock debugger code snippet:
https://github.com/microsoft/vscode-mock-debug/blob/main/src/activateMockDebug.ts#L164

Repro:
- In VSCode, press File -> New Window
- Press F5
- Select "Microsoft Edge Tools" as the environment

Expected:
Some DevTools related debug session will be launched

Actual:
An error toast pops up.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -212,9 +212,26 @@ export function activate(context: vscode.ExtensionContext): void {
         () => {
             telemetryReporter.sendTelemetryEvent('user/buttonPress', { 'VSCode.buttonCode': buttonCode.launchProject });
             LaunchConfigManager.instance.updateLaunchConfig();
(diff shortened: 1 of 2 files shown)


antross (approved):
Looks good for the mentioned repro case. Another similar case I encountered was to do the following without a launch.json:
1. Ctrl+Shift+P
2. Choose "Microsoft Edge Tools: Launch project"

This resulted in a slightly different error message that still occurs with these changes:
> Configuration 'None' is missing in 'launch.json'.

Not sure if that's something we also want to fix here (or want to fix, but in a follow-up). I'm fine either way.

PR #1043, opened 2022-05-20
[Accessibility]: Add verbose label to CDPTargets to indicate existence of action buttons on tab

mliao95:
Issue:
- Screen reader users have no indication that action button items are available on each CDPTarget treeitem
image

Changes:
- Adding verbose label to each tree item. 
    - Items that are expandable/collapsible are identified as CDPTargets and contain the action button list label.
    - Items that are not expandable/collapsible are identified as descriptor tree items and only contain the base label
- The base label is `${this.label}: ${this.description}`.
    - Example:
        - `this.label` = "type"
        - `this.description` = "page"
        - image


antross (approved):
LGTM. Tried it out with Narrator and was able to hear the additional description plus get to the extra buttons as indicated.

PR #1057, opened 2022-05-31
[Feature]: Add context menu option to open html file with Edge DevTools

mliao95:
This PR adds a new context menu item when right clicking HTML files in the VSCode file explorer:
image

Clicking this context menu item will launch a headless version of Edge targeting the page and attach the DevTools.

The end config looks like this:
image

Note that the user does not need a supported `launch.json` to run this workflow.

### DIFF
--- a/package.json
+++ b/package.json
@@ -161,6 +161,11 @@
                 "category": "Microsoft Edge Tools",
                 "enablement": "titleCommandsRegistered",
                 "title": "Toggle CSS mirror editing on/off"
+            },
+            {
+                "command": "vscode-edge-devtools-view.launchHtml",
+                "title": "Open with Edge DevTools",
+                "category": "Microsoft Edge Tools"
             }
         ],
         "configuration": {
@@ -548,7 +553,12 @@
                     "when": "view == vscode-edge-devtools-view.targets && viewItem == cdpTargetProperty",
                     "group": "2_contextMenu"
                 }
-            ]
+            ],
+            "explorer/context": [{
+                "when": "resourceLangId == html",
+                "command": "vscode-edge-devtools-view.launchHtml",
+                "group": "navigation"
+            }]
         },
         "viewsContainers": {
             "activitybar": [
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -37,7 +37,7 @@ import {
     getCSSMirrorContentEnabled,
     setCSSMirrorContentEnabled,
 } from './utils';
-import { LaunchConfigManager } from './launchConfigManager';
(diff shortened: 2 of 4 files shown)


antross:
@mliao95 what do you think about leaving the DevTools closed by default (possibly with some rewording of the context menu entry to avoid implying they'll open automatically)? 

Given we have the button in the screencast to toggle the tools, I think starting users off with the cleaner, screencast-only entrypoint will reduce how much content we're popping in front of them all at once while still letting them get to the DevTools themselves if needed.

package.json
```diff
@@ -37,23 +37,7 @@
...
-        "onLanguage:javascript",
-        "onLanguage:javascriptreact",
-        "onLanguage:json",
-        "onLanguage:jsonc",
-        "onLanguage:less",
-        "onLanguage:sass",
-        "onLanguage:scss",
-        "onLanguage:typescript",
-        "onLanguage:typescriptreact"
+        "onStartupFinished"
```
mliao95, line 40: I'm thinking that we just activate post-VSCode startup.  My thought here is that I want the context menu options available before a user opens an html file.  We already activate on almost all frontend related file types, so the only difference is that we might unnecessarily activate for users who have our extension installed but are working on non-frontend stuff.

Thoughts?

antross: Just curious, what does Live Preview do?

mliao95: Live Preview uses this:
```
"onLanguage:html",
"onWebviewPanel:browserPreview",
"onCommand:livePreview.start",
"onCommand:livePreview.start.preview.atFile",
"onCommand:livePreview.start.debugPreview.atFile",
"onCommand:livePreview.start.internalPreview.atFile",
"onCommand:livePreview.start.externalPreview.atFile",
"onCommand:livePreview.start.externalDebugPreview.atFile"
```

However, Live Server uses `onStartupFinished`.

antross: I don't personally have a strong preference if both approaches work. My main concern is whether removing the explicit `onLanguage:` entries negatively impacts webhint behavior (as I thought we read these entries from the package.json somewhere in code...)

mliao95: I see.  You mean this part?
https://github.com/microsoft/vscode-edge-devtools/blob/main/src/utils.ts#L825

Yeah, maybe this would be a good time to persist the list of supported file types outside of the activation event section of our package.json.

mliao95: Check out the latest commit and let me know what you think:
https://github.com/microsoft/vscode-edge-devtools/pull/1057/commits/705e1ff4b43ce91ff0edd4782b85365247e4dd6f

antross: Hmm... doesn't look like webhint is working for me in the extension with these changes. VS Code pops this error:
image

leotlee (approved):

PR #1087, opened 2022-06-23
Ensure non-undefined document selectors are passed to language client

antross:
The prior approach to splitting entries on `:` resulted in `undefined`
entries being passed to the language client which caused the language
client and server to fail during startup.


mliao95 (approved):
LGTM

PR #1094, opened 2022-06-27
[BugFix] Fix theme fallback being always default systemPreferred

### LINKED ISSUE #1093: Theme auto-detection fallback is always system theme

- OS: Windows 10
- Extension version: 2.0.0

**Describe the bug:**

In #975, the desired behavior of theme auto-detection was:

1. Attempt to map the theme setting (workbench.colorTheme) to a supported theme value
2. Fall back to value based on the active [ColorThemeKind](https://github.com/microsoft/vscode/blob/ebaa253b6d512265c2b3cc99d3174f299c9e4766/src/vscode-dts/vscode.d.ts#L7049-L7068) (enum of Light, Dark, HighContrast, HighContrastLight)
3. Fall back to the system preference

However, when using an unsupported theme, the switch fails to work as intended and always falls through to the default `systemPreferred`.

See related code:
<https://github.com/microsoft/vscode-edge-devtools/blob/6a3d870290dd6de74ea1725867633be4a9048cbb/src/common/settingsProvider.ts#L40-L56>

**Repro steps:**

1. Use a light system theme
2. Configure the workbench.colorTheme setting to a dark theme not listed in the SUPPORTED_THEMES map (e.g. `One Monokai`)
3. Use the extension (e.g. `Launch Edge Headless and attach DevTools`) and see that the dev tools will be light

**Expected behavior:**

The switch with `vscode.window.activeColorTheme.kind` should match a case and set the theme.
If, for example, in the future the enum is extended and there is no match, the default would be `systemPreferred`.

**Additional context:**

This is different from #531 as that predated the code from #975.

maxgubler:
Fixes #1093

The fallback switch used to set `theme` was missing breaks / would always fall through to `default`.

Additionally, updated comments for clarity.

### DIFF
--- a/src/common/settingsProvider.ts
+++ b/src/common/settingsProvider.ts
@@ -3,6 +3,7 @@
 import * as vscode from 'vscode';
 import { SETTINGS_STORE_NAME } from '../utils';
 
+// Map of VS Code theme to Edge DevTools theme
 const SUPPORTED_THEMES = new Map<string, string>([
   ['Default Light+', 'default'],
   ['Visual Studio Light', 'default'],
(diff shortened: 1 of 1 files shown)


captainbrosset:
I'm setting @mliao95 as a reviewer for this, but feel free to change if needed.

PR #1560, opened 2023-06-09
Updating vscode-webhint version to 2.1.10

vidorteg:
Updating vscode-webhint version to `2.1.10` to bring some code fixes and the `css-reflows` hint.


hkal (approved):
LGTM

PR #1804, opened 2023-10-24
User/elormcoch/wsl extension fix

ElormCoch:
Adapts the launch and debug commands based on whether the extension is run in a remote environment.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -243,32 +243,32 @@ export function activate(context: vscode.ExtensionContext): void {
         telemetryReporter.sendTelemetryEvent('contextMenu/launchHtml');
         const edgeDebugConfig = providedHeadlessDebugConfig;
         const devToolsAttachConfig = providedLaunchDevToolsConfig;
-        const { defaultUrl } = getRemoteEndpointSettings();
         if (typeof vscode.env.remoteName === 'undefined') {
             edgeDebugConfig.url = `file://${fileUri.fsPath}`;
             devToolsAttachConfig.url = `file://${fileUri.fsPath}`;
             void vscode.debug.startDebugging(undefined, edgeDebugConfig).then(() => vscode.debug.startDebugging(undefined, devToolsAttachConfig));
         } else {
-            edgeDebugConfig.url = defaultUrl;
-            devToolsAttachConfig.url = defaultUrl;
+            const url = `file://${vscode.env.remoteName}.localhost/${fileUri.authority.split('+')[1]}/${fileUri.fsPath.replace(/\\/g, '/')}`;
+            edgeDebugConfig.url = url;
+            devToolsAttachConfig.url = url;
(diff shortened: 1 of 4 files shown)


src/extension.ts
```diff
@@ -239,20 +239,37 @@ export function activate(context: vscode.ExtensionContext): void {
...
         const devToolsAttachConfig = providedLaunchDevToolsConfig;
-        edgeDebugConfig.url = `file://${fileUri.fsPath}`;
-        devToolsAttachConfig.url = `file://${fileUri.fsPath}`;
-        void vscode.debug.startDebugging(undefined, edgeDebugConfig).then(() => vscode.debug.startDebugging(undefined, devToolsAttachConfig));
+        if (typeof vscode.env.remoteName === 'undefined') {
+            edgeDebugConfig.url = `file://${fileUri.fsPath}`;
+            devToolsAttachConfig.url = `file://${fileUri.fsPath}`;
+            void vscode.debug.startDebugging(undefined, edgeDebugConfig).then(() => vscode.debug.startDebugging(undefined, devToolsAttachConfig));
+        } else {
+            const url = `file://${vscode.env.remoteName}.localhost/${fileUri.authority.split('+')[1]}/${fileUri.fsPath.replace(/\\/g, '/')}`;
```
vidorteg, line 251: could you add a comment with an example of the expected shape of this url?

src/extension.ts
```diff
@@ -239,20 +239,37 @@ export function activate(context: vscode.ExtensionContext): void {
         void setCSSMirrorContentEnabled(context, !cssMirrorContent);
     }));
 
-    context.subscriptions.push(vscode.commands.registerCommand(`${SETTINGS_VIEW_NAME}.launchHtml`, (fileUri: vscode.Uri): void => {
+    context.subscriptions.push(vscode.commands.registerCommand(`${SETTINGS_VIEW_NAME}.launchHtml`, async (fileUri: vscode.Uri): Promise<void> => {
```
vidorteg, line 242: could you also add a test for this scenario in:

https://github.com/microsoft/vscode-edge-devtools/blob/main/test/extension.test.ts

that way we will avoid future regressions.

vidorteg (commented):
Thanks for the fix 👏 @ElormCoch, most of code LGTM,  I left a couple of comments, could you PTAL?

HOLYGHOST1111 (approved):

src/extension.ts
ElormCoch: Done. Thanks!

vidorteg (approved):
Couple of nits but aside from that LGTM 🥳

ElormCoch:
@microsoft-github-policy-service agree company="Microsoft"

vidorteg:
@microsoft-github-policy-service rerun

PR #1993, opened 2024-02-02
Use the same GHA job other repos use, and set permissions

captainbrosset:


vidorteg (approved):
this is extremely helpful, thank you @captainbrosset

PR #2664, opened 2024-12-31
Add instructions for cloning the repository using GitHub CLI

CasP0:


vidorteg:
I'm not sure this is relevant enough for the project to be merged, this seems more intended on how to use git and github repository clone than the product itself. I'm open to revisit this if you believe the information provided is relevant enough.

